PR #10, opened 2023-07-11
Routing functionality

amirsaran3:


esaminu (changes requested):
Per #16 , #17 and #18 those will be the routes on the signer app. Could we update this to match that? Otherwise LGTM

esaminu (approved):
LGTM

Issue #14, opened 2023-07-12
Store metadata about keys and display that to the users for context


DavidM-D:
Create an [authenticated cloud firestore](https://firebase.google.com/docs/firestore/security/rules-conditions#authentication) collection with the structure `users/{userId}/devices` where we can store key to userAgent derived metadata mappings with the schema:

```typescript
{ 
   country: string;
   device: string;
   os: string; 
   publicKeys: string[];
}
```

PR #20, opened 2023-07-21
Initial firebase integration related to devices

hcho112:
This PR contains implementation of firebase related operations. (#14 )
Currently it can do following:
1. Ability to add devices to firestore by go to add device route and click button. (currently using mock uid from test user in test environment)
2. Ability to get list of devices by go to devices route
3. Ability to select and delete devices

There is no style related work involved in this PR, will be added in the future when it is available. 

image

Firestore rules are set so only owner can access to their own collections
```
rules_version = '2';

service cloud.firestore {
  match /databases/{database}/documents {
  	match /users/{userId}/{document=**} {
      allow create, read, update, delete: if request.auth != null && request.auth.uid == userId;
    }
  }
}
```

Owner Accessing their own:
GET:
image
CREATE:
image
DELETE:
image

Unauthenticated user:
GET:
image
CREATE:
image
DELETE:
image

Authenticated but not the owner:
GET:
image
CREATE:
image
DELETE:
image

Result: rules are set correctly so only owner has access to their own. (Others can not see nor create collections)


hcho112:
Updated the test cases

hcho112:
This PR is no longer valid, please refer to this PR https://github.com/near/fast-auth-signer/pull/51

PR #44, opened 2023-09-07
Implement new sign in, sign up flow with latest mpc service end…

hcho112:
This PR contains implementation for adopting new mpc endpoints to handle:
- create new account
- sign in to a new browser
- sign in to a browser that authenticated in the past

### DIFF
--- a/src/components/AuthCallback/AuthCallback.tsx
+++ b/src/components/AuthCallback/AuthCallback.tsx
@@ -1,8 +1,8 @@
 import { KeyPairEd25519 } from '@near-js/crypto';
+import BN from 'bn.js';
 import { isSignInWithEmailLink, signInWithEmailLink } from 'firebase/auth';
 import type { MutableRefObject } from 'react';
-import { useEffect, useRef, useState } from 'react';
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import styled from 'styled-components';
 
@@ -10,6 +10,7 @@ import FastAuthController from '../../lib/controller';
 import { openToast } from '../../lib/Toast';
 import { network, networkId } from '../../utils/config';
 import { firebaseAuth } from '../../utils/firebase';
+import { getUserCredentialsFrpSignature } from '../../utils/mpc-service';
 
 const decodeIfTruthy = (paramVal) => {
   if (paramVal) {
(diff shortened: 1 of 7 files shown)


src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -19,6 +20,173 @@ const decodeIfTruthy = (paramVal) => {
...
+      const parsedUrl = new URL(success_url || window.location.origin);
+      parsedUrl.searchParams.set('account_id', accountIds[0]);
+      parsedUrl.searchParams.set('public_key', public_key_lak);
+      parsedUrl.searchParams.set('all_keys', public_key_lak);
+
+      window.location.replace(parsedUrl.href);
+    },
+  ).catch((error) => {
+    console.log(error);
+    const errorMessages: Record<string, string> = {
```
volovyks, line 171: Code duplication.

src/lib/controller.ts
```diff
@@ -145,6 +160,170 @@ class FastAuthController {
...
+    const CLAIM_SALT = 3177899144 + 0;
+    const keypair = await this.getKey();
+    const signObj = getUserCredentialsFrpSignature({
+      salt: CLAIM_SALT,
+      oidcToken,
+      shouldHashToken: true,
+      keypair,
+    });
+
+    const data = {
```
volovyks, line 176: It should be a good idea to introduce types for requests.

src/lib/controller.ts
```diff
@@ -145,6 +160,170 @@ class FastAuthController {
...
+      method: 'POST',
+      mode: 'cors' as const,
+      body: JSON.stringify(data),
+      headers: new Headers({ 'Content-Type': 'application/json' }),
+    }).then(async (response) => {
+      if (!response.ok) {
+        throw new Error('Unable to claim OIDC token');
+      }
+      const res = await response.json();
+      return res.mpc_signature;
```
volovyks, line 194: To be sure that your token was claimed, you will need to check if this signature is correct using mpc_pk.
Check this code for reference: https://github.com/near/mpc-recovery/blob/develop/integration-tests/src/containers.rs#L838

src/utils/mpc-service.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ConstructSignature } from './types';
+import { SCHEMA } from '@near-js/transactions';
+
+const hashToken = (oidcToken: string): number[] => {
+  const tokenHash = sha256.create();
+  tokenHash.update(oidcToken);
+  return tokenHash.array();
+};
+
+export const getUserCredentialsFrpSignature = ({
```
volovyks, line 14: This function should not have `shouldHashToken` parameter. You never need to hash OIDC Token in `user_credentials` request.

For some reason you are using this function in claim OIDC flow, where we need to hash it, because OIDC Token should not be exposed to the server at that step.
You should create a separate function with hashing for claiming flow.

src/utils/mpc-service.ts
```diff
@@ -0,0 +1,59 @@
...
+  const hash = sha256.create();
+  hash.update(mergeArray);
+
+  return keypair.sign(new Uint8Array(hash.arrayBuffer()));
+};
+
+export const getSignRequestFrpSignature = ({
+  salt, oidcToken, keypair, delegateAction
+}): Signature => {
+  const saltSerialize = serialize(new Map([[Object, { kind: 'struct', fields: [['salt', 'u32']] }]]), ({ salt }));
```
volovyks, line 40: Let's keep these SHEMAs in constants and reuse them. Maybe they are already defined somewhere, not sure.

volovyks (approved):
@hcho112 nice work. None of the comments are blocking. Let's preoretize fast iteration.

src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -19,6 +20,173 @@ const decodeIfTruthy = (paramVal) => {
...
+}) => {
+  const CLAIM_SALT = 3177899144 + 2;
+  const signatureObject = getUserCredentialsFrpSignature({
+    salt: CLAIM_SALT,
+    oidcToken: accessToken,
+    shouldHashToken: false,
+    keypair,
+  });
+  const data = {
+    ...(accountId && accountId.includes('.') ? { near_account_id: accountId } : {}),
```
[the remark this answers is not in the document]
hcho112: (This is existing code before my PR) 
No its more like checking `accountId` existence. However, API doc says that `near_account_id` must be present, so I will remove this part of code

src/lib/controller.ts
hcho112: in JS, it is basically all parsed to string at the moment..

src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -64,88 +232,45 @@ function AuthCallbackPage() {
...
+                publicKeyFak,
+                public_key_lak,
+                contract_id,
+                methodNames,
+                success_url,
+                failure_url,
+                setStatusMessage,
+                email,
+              });
+            }
```
esaminu, line 272: We can put the catch outside and avoid duplication this way
```suggestion
           const accountOperation = isRecovery ? onSignIn : onCreateAccount;

            await accountOperation({
              keypair,
              accountId,
              accessToken: user.accessToken,
              publicKeyFak,
              public_key_lak,
              contract_id,
              methodNames,
              success_url,
              failure_url,
              setStatusMessage,
              email,
            }).catch((error) => {
              console.log(error);
              const errorMessages: Record<string, string> = {
                'auth/expired-action-code': 'Link expired, please try again.',
                'auth/invalid-action-code': 'Link expired, please try again.',
                'auth/invalid-email':       'Invalid email address.',
                'auth/user-disabled':       'User disabled',
                'auth/missing-email':       'No email found, please try again.',
              };
              const message = errorMessages[error.code] || error.message;
              const parsedUrl = new URL(failure_url || success_url || window.location.origin);
              parsedUrl.searchParams.set('code', error.code);
              parsedUrl.searchParams.set('reason', message);
              window.location.replace(parsedUrl.href);
              openToast({
                type:  'ERROR',
                title: message,
              });
            });
```

esaminu (approved):
LGTM minor refactor but no blockers ✅

hcho112:
Please hold on merging this PR, I have another PR that contains all the changes in this PR + firestore integration +change request from this PR all together.

DavidM-D:
@hcho112 where is that PR?

hcho112:
@DavidM-D It's in my local branch, I will push the PR when it is ready

PR #51, opened 2023-09-19
Integration on new mpc service + firestore + relayer

hcho112:
This PR contains following implementation:
- Integration on new mpc endpoint
- Integration on firestore
- Integration on relayer
- Able to create new Account with above
- Able to signin with existing account with above
- Able to delete existing keys and continue signin with above

**There are more work need on following**:
- UI/messaging around the is the meaning of exceeded maximum keys
- Lack of verification on mpc_service against `claim_oidc` token
- Lack of UI on device tab
- Pull out relayer, firestore configuration and make it customizable

Firestore permission setting:
- User can perform GET, POST, UPDATE, DELETE on their own collection only

To test this branch locally, need to run [near discovery](https://github.com/near/near-discovery) locally by pull down and run following:
```
git checkout fast-auth-iframe-wallet
pnpm install && pnpm dev -p 3001
```
Then checkout to this branch and run following:
```
yarn && yarn start
```

High level description on authentication flow:

On create account:
-> create new account through firebase 
-> hit `/new_account` endpoint on mpc service
-> create new firestore record that contains one FAK and one LAK
-> back to app

On Signin to a browser that has setup passKey
-> use webauthn key to get recovery key
-> attempt to create new LAK and delete old LAK associated to webauthn key
(If fail due to too much key, it will redirect to `/devices` endpoint, will be described later)
-> update firestore collection with only new LAK
-> back to app

On Signin to a browser that has no passkey setup yet
-> use webauthn to check if browser has key
-> if no key, create a new webAuthN key and LAK
(If fail due to too much key, it will redirect to `/devices` endpoint, will be described later)
-> create new firestore record that contains one FAK and one LAK
-> back to app

On Signin to a browser that has passkey but no longer on chain (deleted by `/devices` page)
-> use webauthn to check if browser has key
-> if it has key, but doesn't exist on chain, create a new webAuthN key and LAK
(If fail due to too much key, it will redirect to `/devices` endpoint, will be described later)
-> create new firestore record that contains one FAK and one LAK
-> back to app

On Redirected to a `/devices` page
-> Get list of device records from firestore
-> User select and delete firestore record
-> Also delete FAK and LAK associated to deleted firestore record
-> create new WebAuthN FAK and LAK
-> create new firestore record that contains one FAK and one LAK
-> back to app


src/lib/useAuthState.ts
```diff
@@ -1,14 +1,19 @@
...
+import { useEffect, useMemo, useState } from 'react';
+import { useSearchParams } from 'react-router-dom/dist';
 
-import FastAuthController from "./controller";
-import { network, networkId } from "../utils/config";
+import FastAuthController from './controller';
+import { network, networkId } from '../utils/config';
 
-export const useAuthState = () => {
-  const [authenticated, setAuthenticated] = useState(undefined);
```
esaminu, line 11: This needs to start as `undefined` because we need to identify when the user has not authenticated yet vs failed authentication in some cases

hcho112: On that exact reason, I used `controllerState` which already does it inside the code and export it:

https://github.com/near/fast-auth-signer/pull/51/files#diff-e32098d15337091eba6e38cf673f428139718632e81710d1e180d58836cae11bR12?plain=1

hcho112: Eg.

https://github.com/near/fast-auth-signer/pull/51/files#diff-8e391524757018ec63bd655f87614c8e65b6457ff6b5150c710bd026796da112R111

https://github.com/near/fast-auth-signer/pull/51/files#diff-8e391524757018ec63bd655f87614c8e65b6457ff6b5150c710bd026796da112R125

hcho112: so I have updated the logic on Sign component

esaminu: `controllerState` could be true (key is retrieved from keyStore) while the user still has not authenticated yet right (e.g. they are still selecting an account from the list). I don't mind exposing `controllerState` from this hook but I prefer for `authenticated` to have a loading state as well and for components to use that to determine if the user has fully authenticated instead of `controllerState` since `controllerState` is one of the variables that determines whether `authenticated` is `true` or not

esaminu (approved):
LGTM ✅

PR #52, opened 2023-09-19
login page

DamirSQA:


src/components/AddDevice/AddDevice.tsx
```diff
@@ -57,94 +74,135 @@ function SignInPage() {
...
-    const success_url = searchParams.get('success_url');
-    const failure_url = searchParams.get('failure_url');
-    const public_key =  searchParams.get('public_key');
-    const contract_id = searchParams.get('contract_id');
-    const methodNames = searchParams.get('methodNames');
+  const user = firebaseAuth.currentUser;
+
+  if (user !== null) {
+    user.providerData.forEach((profile) => {
+      console.log('Sign-in provider: ' + profile.providerId);
```
amirsaran3, line 85: remove console.log

src/components/Login/Login.tsx
```diff
@@ -4,6 +4,12 @@ import { useLocation, useNavigate, useSearchParams } from 'react-router-dom';
 import { Button } from '../../lib/Button';
```
esaminu, line 4: looks like the `isSignedIn` and the related `useEffect` in this file are no longer required. Also, can you confirm that all redirects are working as expected whether `basePath` is an empty string or has a value e.g. `fastauth`

esaminu (approved):
LGTM

PR #56, opened 2023-09-22
Hide instructions on the bug report template

ewiner:
I find that putting bug reporting instructions inside comment blocks (`<!-- comment -->`) helps avoid having those instructions littered over the actual bug reports, as people rarely delete that text.

I'm just being pedantic here; if you prefer it the way it is, feel free to close this PR.


trechriron (approved):
This is a great idea. Thanks!

PR #72, opened 2023-10-03
Don't add FAK when no biometrics

amirsaran3:
Added check to see if biometrics is supported on browser before adding FAK.


DamirSQA (approved):

esaminu:
@amirsaran3 The flow will still fail if webauthn is not supported, we need to make sure that the whole flow works end to end on browsers where webauthn is not supported (e.g. firefox). For example, [we currently still throw in `handleCreateAccount`](https://github.com/near/fast-auth-signer/blob/main/src/components/AddDevice/AddDevice.tsx#L75-L77). Please test the flow end to end and make sure it works the same in webauthn supported browsers and still works in non webauthn supported browser just without a keypair added to the keystore or a username added to localStorage

esaminu:
Closing in favor of #90 thanks @amirsaran3 !

PR #74, opened 2023-10-04
fix: removing outdated controllerState on AuthIndicator

hcho112:
This PR contains bug fix for removing unsupported `controllerState` from AuthIndicator component


esaminu (approved):
LGTM

PR #79, opened 2023-10-09
Added lint workflow

DamirSQA:
Added PR workflow that checks for lint errors and blocks merging if there's lint issues present. Lint fixes included.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,8 @@ module.exports = {
     'react/jsx-filename-extension':      'off',
     'react/jsx-props-no-spreading':      'off',
     'import/no-extraneous-dependencies': 'off',
+    'import/prefer-default-export':      'off',
+    'react/no-array-index-key':          'off',
     'prefer-destructuring':                [
       'error', {
         VariableDeclarator:   {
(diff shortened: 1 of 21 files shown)


.eslintrc.js
```diff
@@ -22,6 +22,8 @@ module.exports = {
     'react/jsx-filename-extension':      'off',
     'react/jsx-props-no-spreading':      'off',
     'import/no-extraneous-dependencies': 'off',
+    'import/prefer-default-export':      'off',
+    'react/no-array-index-key':          'off',
```
MaximusHaximus, line 26: :hocho: using index for react keys is an antipattern we want to avoid that can lead to hard to diagnose bugs when unrelated code is refactored or data lifecycles change -- let's keep this on please

MaximusHaximus (changes requested):
Good idea, let's get this landed, but please keep our react 'no array index as key' rule enabled and fix the code occurences that break this rule instead of disabling it.

.eslintrc.js
MaximusHaximus: Looking at `Sign.tsx`, there really is no stable ID we can use there and the transaction data should not change, so in that particular case it's fine to disable the eslint rule for that one place; creating UUIDs here wouldn't really serve any purpose. The main thing is just that don't want to disable the rule entirely for the entire codebase, as this type of eslint error serves as a 'friendly reminder' for a developer who has just used an array index as a key to look carefully at the data they are working with and be absolutely sure it's OK to use an array index as the key in the place where they are doing so.

DamirSQA: Updated as suggested, thanks.

PR #88, opened 2023-10-12
Redirect instead of prompt if no email in local storage

amirsaran3:
- Instead of prompting user to confirm email now it redirects the user to `failure_url` with message: "Please use the same device and browser to verify your email"


esaminu (approved):
LGTM

PR #90, opened 2023-10-14
feat: add support for non webauthn browsers

### LINKED ISSUE #87: [Defect] FastAuth has stopped working on Firefox
Screenshot 2023-10-11 at 21 50 01

When trying to Sign in to FastAuth in Firefox in Mac, after the first redirect you get a dom error whenever you press the Sign In button.

```
DOMException: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission. AddDevice.tsx:154
```

esaminu:
Closes #87

### DIFF
--- a/src/components/AddDevice/AddDevice.tsx
+++ b/src/components/AddDevice/AddDevice.tsx
@@ -1,4 +1,4 @@
-import { createKey } from '@near-js/biometric-ed25519/lib';
+import { createKey, isPassKeyAvailable } from '@near-js/biometric-ed25519/lib';
 import BN from 'bn.js';
 import { sendSignInLinkToEmail } from 'firebase/auth';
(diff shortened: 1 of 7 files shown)


hcho112 (changes requested):
I have conducted the test against this part of code on Firefox and unfortunately it doesn't work. As part of communicating to relayer, we are expected to have `keypair` that will be used to create `signature` (https://github.com/near/mpc-recovery)

This PR is currently not working because it is missing `keypair` on firefox environment. 

I will work on the fix and push to this PR shortly

hcho112:
@esaminu I'm still working on the flow of when max key reached path on top of this PR

hcho112 (approved):
LGTM, tested multiple times on Chrome and Firefox browsers

hcho112 (approved):
LGTM!

PR #92, opened 2023-10-19
feat: Support on devices page with ff browser

hcho112:
This PR contains implementation to support non-webauthn browser (eg. Firefox) 

Tested following flow:
- Create account from Firefox, Sign out and Sign back in with same browser
- Use above account to sign in from Chrome browser (cross browser testing)
- Use account created on Chrome to sign in on Firefox
- Hit max key flow and on devices, delete keys and proceed to signin

### DIFF
--- a/src/components/Devices/Devices.tsx
+++ b/src/components/Devices/Devices.tsx
@@ -1,16 +1,12 @@
-import { KeyPairEd25519 } from '@near-js/crypto';
+import { isPassKeyAvailable } from '@near-js/biometric-ed25519';
 import React, { useEffect, useMemo, useState } from 'react';
 import { useSearchParams, useNavigate } from 'react-router-dom';
 import { styled } from 'styled-components';
 
-import FastAuthController from '../../lib/controller';
 import FirestoreController from '../../lib/firestoreController';
 import { useAuthState } from '../../lib/useAuthState';
 import { decodeIfTruthy } from '../../utils';
-import { networkId } from '../../utils/config';
 import { onSignIn } from '../AuthCallback/AuthCallback';
-import { getDomain } from '../../utils/firebase';
-import { isPassKeyAvailable } from '@near-js/biometric-ed25519';
 
 const Title = styled.h1`
   padding-bottom: 20px;
@@ -54,8 +50,8 @@ function Devices() {
   const controller = useMemo(() => new FirestoreController(), []);
   const [searchParams] = useSearchParams();
   const navigate = useNavigate();
-  const public_key_lak = decodeURIComponent(searchParams.get('public_key_lak'));
-  const publicKeyFak = decodeURIComponent(searchParams.get('publicKeyFak'));
+  const public_key_lak = decodeIfTruthy(searchParams.get('public_key_lak')) || decodeIfTruthy(searchParams.get('public_key'));
+  const publicKeyFak = decodeIfTruthy(searchParams.get('publicKeyFak'));
   const { authenticated } = useAuthState();
 
   const onClick = (id) => {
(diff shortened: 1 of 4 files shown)


src/components/Devices/Devices.tsx
```diff
@@ -69,26 +65,17 @@ function Devices() {
...
       setCollections(deviceCollections);
     };
-    if (authenticated !== 'loading') {
+    if (controller.getUserOidcToken()) {
       getCollection();
+    } else {
+      navigate({
+        pathname: '/login',
+        search:   searchParams.toString(),
+      });
```
hcho112, line 82: Have you encountered a situation where you had to redirect to login state?

src/components/Devices/Devices.tsx
```diff
@@ -54,8 +50,8 @@ function Devices() {
   const controller = useMemo(() => new FirestoreController(), []);
   const [searchParams] = useSearchParams();
   const navigate = useNavigate();
-  const public_key_lak = decodeURIComponent(searchParams.get('public_key_lak'));
-  const publicKeyFak = decodeURIComponent(searchParams.get('publicKeyFak'));
+  const public_key_lak = decodeIfTruthy(searchParams.get('public_key_lak')) || decodeIfTruthy(searchParams.get('public_key'));
+  const publicKeyFak = decodeIfTruthy(searchParams.get('publicKeyFak'));
```
hcho112, line 53: good change 👍

src/components/Devices/Devices.tsx
```diff
@@ -103,45 +90,35 @@ function Devices() {
...
+        if (contract_id && public_key_lak) {
+          setIsAddingKey(true);
+
+          const email = decodeIfTruthy(searchParams.get('email'));
+          const methodNames = decodeIfTruthy(searchParams.get('methodNames'));
+          const success_url = decodeIfTruthy(searchParams.get('success_url'));
+          const oidcToken = await controller.getUserOidcToken();
+          await onSignIn({
+            accessToken:      oidcToken,
+            publicKeyFak,
```
hcho112, line 113: hmm after removing `getPublicKeyFak` have you tried to test create account, signin and multiple key flow on firefox?

src/components/Devices/Devices.tsx
```diff
@@ -71,12 +70,18 @@ function Devices() {
     if (controller.getUserOidcToken()) {
       getCollection();
     } else {
-      navigate({
-        pathname: '/login',
-        search:   searchParams.toString(),
+      (new Promise((resolve) => { setTimeout(resolve, 5000); })).then(controller.getUserOidcToken).then((token) => {
```
hcho112, line 73: I assume that this part of code is delaying the checking token by 5 seconds. (I think this may be too long?) 

Also maybe we can rewrite this part as recursive check, eg every 1 second we check if token is ready for 3 times and if within 3 tries if it exist, we stop next recursive call and execute `getCollection` or if it failed to check after 3 tries, we execute redirect logic? 

But for the time being, maybe we go with what we have

src/components/Devices/Devices.tsx
esaminu: This should happen if you navigate to the devices page directly meaning currently the devices page requires an auth token and can't just be used with FAK. We should fix that soon

src/components/Devices/Devices.tsx
esaminu: Yes it works on my side

src/components/Devices/Devices.tsx
esaminu: We can use interval checking or exponential backoff type behavior here. Feel free to open an issue regarding this

hcho112 (commented):
Completed multiple test against Chrome and Firefox browser

PR #102, opened 2023-10-30
fix: throw if no account exists

esaminu:
Throw in the following cases:

1. Login attempt with non existing account on firebase 
2. Account does not exist on chain due to acc creation previously failed
3. Account can't be looked up due to indexer falling behind

### DIFF
--- a/src/components/AddDevice/AddDevice.tsx
+++ b/src/components/AddDevice/AddDevice.tsx
@@ -1,6 +1,6 @@
 import { createKey, isPassKeyAvailable } from '@near-js/biometric-ed25519/lib';
 import BN from 'bn.js';
-import { sendSignInLinkToEmail } from 'firebase/auth';
+import { fetchSignInMethodsForEmail, sendSignInLinkToEmail } from 'firebase/auth';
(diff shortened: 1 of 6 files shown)


src/components/AddDevice/AddDevice.tsx
```diff
@@ -162,6 +166,11 @@ function SignInPage() {
       navigate(`/verify-email?${newSearchParams.toString()}#${hashParams.toString()}`);
     } catch (error: any) {
       console.log(error);
+      const { message } = error;
+      const parsedUrl = new URL(failure_url || success_url || window.location.origin + (basePath ? `/${basePath}` : ''));
+      parsedUrl.searchParams.set('code', error.code);
+      parsedUrl.searchParams.set('reason', message);
+      window.location.replace(parsedUrl.href);
```
hcho112, line 173: I see this part of code gets repeated multiple times, is it possible if we pull this into a util function and repeat on other areas?

hcho112 (commented):
Rest looks good. Just comment on refactoring.

PR #110, opened 2023-11-09
Add playwright test app

### LINKED ISSUE #109: Create app to facilitate Playwright tests via an embedded iframe
In order to integration test our signer, we need to be able to use it in the same way it is used in e.g. `near-discovery`, but without all of the other complexities of that application being brought into our tests.

We will create a small app that embeds `near-fast-auth-signer` in an iframe that is designed specifically for our testing purposes.

MaximusHaximus:
- Facilitates testing fast-auth-signer without needing to use near-discovery to load the app

Closes #109


esaminu (approved):
LGTM

PR #114, opened 2023-11-23
Supporting non webauthn browser with FAK action

hcho112:
This PR contains implementation to support FAK actions on non webauthn browser. 

before
after

Right now, if I were to execute non-authorised FAK action and pass it to `/sign` on mpc recovery service, it returns with above.

### DIFF
--- a/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
+++ b/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
@@ -2,7 +2,9 @@ import { yupResolver } from '@hookform/resolvers/yup';
 import { captureException } from '@sentry/react';
 import BN from 'bn.js';
 import { sendSignInLinkToEmail } from 'firebase/auth';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState
+} from 'react';
 import { useForm } from 'react-hook-form';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import styled from 'styled-components';
@@ -259,7 +261,7 @@ function SignInPage() {
     if (requireVerifyEmail) return;
 
     handleAuthCallback();
-  }, [addDevice, authenticated, navigate, searchParams]);
+  }, [addDevice, authenticated, navigate, searchParams, requireVerifyEmail]);
 
   if (authenticated === true && !requireVerifyEmail) {
     return renderRedirectButton ? (
--- a/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
+++ b/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
@@ -168,13 +168,18 @@ function SignInPage() {
         const publicKeyFak = await window.fastAuthController.getPublicKey();
         const existingDevice = await window.firestoreController.getDeviceCollection(publicKeyFak);
         const existingDeviceLakKey = existingDevice?.publicKeys?.filter((key) => key !== publicKeyFak)[0];
+        const user = firebaseAuth.currentUser;
+        // @ts-ignore
+        const oidcToken = user.accessToken;
+        const recoveryPK = await window.fastAuthController.getUserCredential(oidcToken);
+
         // if given lak key is already attached to webAuthN public key, no need to add it again
         const noNeedToAddKey = existingDeviceLakKey === public_key;
         if (noNeedToAddKey) {
           const parsedUrl = new URL(success_url || window.location.origin + (basePath ? `/${basePath}` : ''));
           parsedUrl.searchParams.set('account_id', window.fastAuthController.getAccountId());
           parsedUrl.searchParams.set('public_key', public_key);
-          parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak].join(','));
+          parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak, recoveryPK].join(','));
 
           if (inIframe()) {
             setRenderRedirectButton(parsedUrl.href);
@@ -197,12 +202,10 @@ function SignInPage() {
           }
 
           // Add device
-          const user = firebaseAuth.currentUser;
           window.firestoreController.updateUser({
             userUid:   user.uid,
             // User type is missing accessToken but it exist
-            // @ts-ignore
-            oidcToken: user.accessToken,
+            oidcToken,
           });
 
           // Since FAK is already added, we only add LAK
@@ -215,7 +218,7 @@ function SignInPage() {
               const parsedUrl = new URL(success_url || window.location.origin + (basePath ? `/${basePath}` : ''));
               parsedUrl.searchParams.set('account_id', window.fastAuthController.getAccountId());
               parsedUrl.searchParams.set('public_key', public_key);
-              parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak].join(','));
+              parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak, recoveryPK].join(','));
               window.parent.postMessage({
                 type:   'method',
                 method: 'query',
@@ -223,8 +226,8 @@ function SignInPage() {
                 params: {
                   request_type: 'complete_sign_in',
                   publicKey:    public_key,
-                  allKeys:      [public_key, publicKeyFak].join(','),
-                  accountId:    window.fastAuthController.getAccountId()
+                  allKeys:      [public_key, publicKeyFak, recoveryPK].join(','),
+                  accountId:    (window as any).fastAuthController.getAccountId()
                 }
               }, '*');
               if (inIframe()) {
--- a/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
+++ b/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
@@ -93,9 +93,6 @@ function SignInPage() {
   const skipGetKey = decodeIfTruthy(searchParams.get('skipGetKey'));
   const { authenticated } = useAuthState(skipGetKey);
   const [renderRedirectButton, setRenderRedirectButton] = useState('');
-  // if requireVerifyEmail is set, user need to verify email first
-  const requireVerifyEmail = useMemo(() => window.localStorage.getItem('requireVerifyEmail'), []);
-
   if (!window.firestoreController) {
     window.firestoreController = new FirestoreController();
   }
@@ -164,18 +161,20 @@ function SignInPage() {
       const methodNames = decodeIfTruthy(searchParams.get('methodNames'));
 
       const email = decodeIfTruthy(searchParams.get('email'));
-      if (authenticated === true && isFirestoreReady) {
+      const isPasskeySupported = await isPassKeyAvailable();
+      const user = firebaseAuth.currentUser;
+      const firebaseAuthInvalid = authenticated === true && !isPasskeySupported && user?.email !== email;
+      if (authenticated === true && isFirestoreReady && !firebaseAuthInvalid) {
         if (!public_key || !contract_id) {
           window.location.replace(success_url || window.location.origin + (basePath ? `/${basePath}` : ''));
           return;
         }
-        const isPasskeySupported = await isPassKeyAvailable();
         const publicKeyFak = isPasskeySupported ? await window.fastAuthController.getPublicKey() : '';
         const existingDevice = isPasskeySupported
           ? await window.firestoreController.getDeviceCollection(publicKeyFak)
           : null;
         const existingDeviceLakKey = existingDevice?.publicKeys?.filter((key) => key !== publicKeyFak)[0];
-        const user = firebaseAuth.currentUser;
+
         // @ts-ignore
         const oidcToken = user.accessToken;
         const recoveryPK = await window.fastAuthController.getUserCredential(oidcToken);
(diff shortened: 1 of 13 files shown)


packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -145,8 +155,41 @@ class FastAuthController {
   }
 
   async signDelegateAction({ receiverId, actions, signerId }) {
-    this.assertValidSigner(signerId);
+    const isSigned = await this.isSignedIn();
+    if (this.accountId !== signerId) {
+      this.setAccountId(signerId);
+      this.localStore = new keyStores.BrowserLocalStorageKeyStore();
+    }
```
esaminu, line 162: What is the purpose of this code block? AFAICT the only time we override the `accountId` is when there is no `accountId` in the controller and in that case we derive it from the oidc token and set it

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -22,21 +23,24 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
   }, []);
 
   const [controllerState, setControllerState] = useState<'loading' | boolean>('loading');
+  const [isPassKeySupported, setIsPassKeySupported] = useState<boolean>(null);
+  isPassKeyAvailable().then((isAvailable) => setIsPassKeySupported(isAvailable));
```
esaminu, line 27: Should we run this once on mount in a `useEffect`?

packages/near-fast-auth-signer/src/lib/controller.ts
hcho112: I have add them in because in some cases,
controller is initialised with:
https://github.com/near/fast-auth-signer/blob/1cd623db107e12109de7077aa7cf0c7dd009f164/packages/near-fast-auth-signer/src/App.tsx#L22-L25
where accountId is wrong.

esaminu (dismissed):
LGTM, lets fix the lint error and land this

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -201,13 +201,18 @@ function SignInPage() {
         const publicKeyFak = await window.fastAuthController.getPublicKey();
         const existingDevice = await window.firestoreController.getDeviceCollection(publicKeyFak);
         const existingDeviceLakKey = existingDevice?.publicKeys?.filter((key) => key !== publicKeyFak)[0];
+        const user = firebaseAuth.currentUser;
+        // @ts-ignore
+        const oidcToken = user.accessToken;
```
Pessina, line 206: Instead of disabling the typescript can we use the `user.getIdToken()` ? 

I logged it a few times on different scenarios and looks like it returns the same token as the accessToken. 

I tried to find the docs for the accessToken and check if it's idd the same as `user.getIdToken()` but I can't find the docs for accessToken

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -248,15 +251,15 @@ function SignInPage() {
               const parsedUrl = new URL(success_url || window.location.origin + (basePath ? `/${basePath}` : ''));
               parsedUrl.searchParams.set('account_id', (window as any).fastAuthController.getAccountId());
               parsedUrl.searchParams.set('public_key', public_key);
-              parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak].join(','));
+              parsedUrl.searchParams.set('all_keys', [public_key, publicKeyFak, recoveryPK].join(','));
```
Pessina, line 254: I see this line is duplicated in 215, can we avoid duplicating it and re-use the parsedUrl defined above?

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -146,13 +155,31 @@ class FastAuthController {
...
+      const account = new Account(this.connection, this.accountId);
+      return account.signedDelegate({
+        actions,
+        blockHeightTtl: 60,
+        receiverId,
+      });
+    } catch {
+      // fallback, non webAuthN supported browser
+      // @ts-ignore
+      const oidcToken =  firebaseAuth.currentUser.accessToken;
```
Pessina, line 169: Instead of disabling the typescript can we use the user.getIdToken() ?

I logged it a few times on different scenarios and looks like it returns the same token as the accessToken.

I tried to find the docs for the accessToken and check if it's idd the same as user.getIdToken() but I can't find the docs for accessToken

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -146,13 +155,31 @@ class FastAuthController {
...
+      const recoveryPK = await this.getUserCredential(oidcToken);
+      // make sure to handle failure, (eg token expired) if fail, redirect to failure_url
+      return this.createSignedDelegateWithRecoveryKey({
+        oidcToken,
+        accountId: this.accountId,
+        actions,
+        recoveryPK,
+      }).catch((err) => {
+        console.log(err);
```
Pessina, line 178: Just for curiosity, why we log the errors on user console?

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -146,13 +155,31 @@ class FastAuthController {
...
+    try {
+      // webAuthN supported browser
+      const account = new Account(this.connection, this.accountId);
+      return account.signedDelegate({
+        actions,
+        blockHeightTtl: 60,
+        receiverId,
+      });
+    } catch {
+      // fallback, non webAuthN supported browser
```
Pessina, line 167: Just wondering, instead of using try/catch for a known scenario. Can we check if the browser supports webAuthN and then chose the code to run?

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -269,15 +303,16 @@ class FastAuthController {
...
+  async createSignedDelegateWithRecoveryKey({
     oidcToken,
     accountId,
     recoveryPK,
     actions,
   }) {
     const GET_SIGNATURE_SALT = CLAIM + 3;
     const GET_USER_SALT = CLAIM + 2;
-    const localKey = await this.getKey(`oidc_keypair_${oidcToken}`);
+    const localKey = await this.getKey(`oidc_keypair_${oidcToken}`) || await this.getLocalStoreKey(`oidc_keypair_${oidcToken}`);
```
Pessina, line 314: Duplicated line. If this will be a common scenario does it makes sense create a method for this get?

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -199,6 +226,7 @@ class FastAuthController {
     if (!keypair) {
       keypair = KeyPair.fromRandom('ED25519');
       await this.keyStore.setKey(this.networkId, `oidc_keypair_${oidcToken}`, keypair);
+      await this.localStore.setKey(this.networkId, `oidc_keypair_${oidcToken}`, keypair);
```
Pessina, line 229: Just for curiosity, why we need the keyStore and localStore? Isn't bad to have to keep to places in sync and have two sources of truth?

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -22,21 +23,28 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
   }, []);
 
   const [controllerState, setControllerState] = useState<'loading' | boolean>('loading');
+  const [isPassKeySupported, setIsPassKeySupported] = useState<boolean>(null);
+
+  useEffect(() => {
```
Pessina, line 28: Do we need an extra useEffect and useState? 

I think we could to this fetch inside the main useEffect

Something like this

```
useEffect(() => {
    const isPasskeyAvailable = await isPasskeyAvailable()

    if (isPassKeySupported === null) return;
    if (skipGetKeys) {
      setAuthenticated(false);
      setControllerState(false);
    } else if (controllerState !== false) {
.
.
.
```

That way we can remove the extra useEffect and useState

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -63,8 +71,31 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
...
           }
         }).catch(() => setAuthenticated(false));
+    } else {
+      checkFirestoreReady().then(async (isReady) => {
+        if (isReady) {
+          // @ts-ignore
+          const oidcToken = firebaseAuth.currentUser.accessToken;
+          if (window.fastAuthController.getLocalStoreKey(`oidc_keypair_${oidcToken}`)) {
+            const recoveryPK = await window.fastAuthController.getUserCredential(oidcToken);
+            const accountIds = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${recoveryPK}/accounts`)
```
Pessina, line 81: On this scenario accountIds will be type any, no? 

Can we defined the API call on a separated file, handle the error and properly type it's return? 

I'm placing the API calls on this file: https://github.com/near/fast-auth-signer/blob/main/packages/near-fast-auth-signer/src/api/index.ts

But we can also discus how we can organize it. 

I just don't like the idea of spreading API calls across the codebase and not typing it's return type.

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -63,8 +71,31 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
...
+              .then((res) => res.json())
+              .catch((err) => {
+                console.log(err);
+                throw new Error('Unable to retrieve account Id');
+              });
+            (window as any).fastAuthController = new FastAuthController({
```
Pessina, line 87: Do we need the any here? I guess the fastAuthController it's defined on the global ts file, no?

Pessina (changes requested):

packages/near-fast-auth-signer/src/lib/useAuthState.ts
esaminu: We can't make a useEffect callback async directly so the only other option would be to wrap everything else in a promise or iife. Stylistically the current solution is slightly better than that imo

packages/near-fast-auth-signer/src/lib/useAuthState.ts
esaminu: Agreed, let's throw an issue on the board to collect all of these and organize into service files wdyt? Unfortunately we are only now in a position to address tech debt

packages/near-fast-auth-signer/src/lib/useAuthState.ts
Pessina: I'm sorry, idd. 

I mean: 

```
useEffect(() => {
  const exampleName = async () => {
    const isPasskeyAvailable = await isPasskeyAvailable();

    if (isPasskeyAvailable === null) return;

    if (skipGetKeys) {
      setAuthenticated(false);
      setControllerState(false);
    } else if (controllerState !== false) {
      // ...
    }
  };

  exampleName();
}, [...deps]); 
```

esaminu: With the new changes on main it now makes sense to await `isPasskeyAvailable` so will make that change when I rebase this

packages/near-fast-auth-signer/src/lib/useAuthState.ts
Pessina: Yeah I understand and agree 100%, let's start creating tickets for the tech debt when we face them.

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -59,6 +62,28 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
         } catch {
           setAuthenticated(false);
         }
+      } else {
+        checkFirestoreReady().then(async (isReady) => {
+          if (isReady) {
+            const oidcToken = firebaseAuth.currentUser.getIdToken();
```
Pessina, line 68: I think we need to await here

Pessina (changes requested):

packages/near-fast-auth-signer/src/lib/controller.ts
esaminu: The specific condition we are checking for is whether there is a key in the controller. The controller is indifferent to whether the browser supports webauthn or not, it just tries to sign the delegate locally and if that fails, it falls back to using the recovery service. We could possibly be more explicit by checking if we have a key in the local keystore before trying to sign here but the current logic is fine too wdyt

packages/near-fast-auth-signer/src/lib/controller.ts
esaminu: Good point. The design of this implementation needs both ephemeral in memory keystore and the local storage keystore that persists between sessions. The in memory keystore will store FAKs after retrieving from webauthn and the local storage keystore will store a permanent mapping of oidc token to claimed keypair so that we can sign things with the recovery service permanently in non webauthn browsers. I believe it's possible for us to move the oidctoken <> keypair mapping permanently to localstorage removing it from in memory storage since we are storing it there now anyway. Will do some further testing and circle back on this

packages/near-fast-auth-signer/src/lib/controller.ts
esaminu: Dependent on above discussion re multiple keystores so leaving as is for now

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -59,6 +62,28 @@ export const useAuthState = (skipGetKeys = false): AuthState => {
         } catch {
           setAuthenticated(false);
         }
+      } else {
+        checkFirestoreReady().then(async (isReady) => {
+          if (isReady) {
+            const oidcToken = await firebaseAuth.currentUser.getIdToken();
+            if (window.fastAuthController.getLocalStoreKey(`oidc_keypair_${oidcToken}`)) {
+              const recoveryPK = await window.fastAuthController.getUserCredential(oidcToken);
+              const accountIds = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${recoveryPK}/accounts`)
```
Pessina, line 71: Can we re-use the function?

https://github.com/near/fast-auth-signer/blob/d5c66ff774a73375784f6fdf6c412afb7007cde7/packages/near-fast-auth-signer/src/api/index.ts#L15

Pessina (approved):

esaminu (approved):
LGTM

Pessina (approved):

PR #115, opened 2023-11-27
Sentry config improvements

hcho112:
This PR contains several improvements from sentry
- Hide key related query params
- Only enable on mainnet
- Avoid config setup if no DSN is passed 
- Reduce sample rate to 10%
- Hide breadcrumb to avoid interaction with other parties


esaminu (approved):
LGTM

PR #118, opened 2023-11-28
Data-test-ids for playwright testing

trechriron:
Added data-test-ids to form fields and buttons for playwright testing.


esaminu (approved):
LGTM

PR #132, opened 2023-12-22
Re-use Input component on Login and AddDevice screen

Pessina:


packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -265,7 +265,7 @@ function CreateAccount() {
           error={errors?.username?.message}
           subText="Use a suggested ID or customize your own"
           autoComplete="webauthn username"
-          right=".near"
+          right={network.fastAuth.accountIdSuffix}
```
DamirSQA, line 268: Could you please add dot , should be present before network suffix,

DamirSQA:
Hi @Pessina , I was testing the PR, and noticed that bug is introduced when selecting email domains, and it would be great if it can be addressed. I've made [the recording](https://www.loom.com/share/95724839071849aab30712e1d7c59d03) of the issue, so you can check that out. Thanks!

DamirSQA (approved):

PR #136, opened 2024-01-01
Replace RPC call with near-api-js when checking account existence

Pessina:


packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -110,14 +71,11 @@ const schema = yup.object().shape({
...
       async (username, context) => {
-        if (username) {
-          const isAvailable = await checkIsAccountAvailable(username);
-          if (!isAvailable) {
-            return context.createError({
-              message: `${username}.${network.fastAuth.accountIdSuffix} is taken, try something else.`,
-              path:    context.path
-            });
-          }
+        if (username && await window.fastAuthController.accountExist(username)) {
```
hcho112, line 74: I believe this part of logic has been re-written recently

hcho112:
Is this PR still relevant? @Pessina

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
Pessina: By what I checked, we only changed the e-mail validation logic. The username validation remains the same.

Pessina:
> Is this PR still relevant? @Pessina

I believe yes. 

We will replace the RPC call and rely on the already implemented function in near-api-js simplifying the code and saving us some lines. 

What do you think?

PR #139, opened 2024-01-02
Re-use fetchAccountIds on all necessary places

Pessina:


packages/near-fast-auth-signer/src/lib/Input/Input.tsx
```diff
@@ -7,9 +7,9 @@ import SuccessSvg from '../../components/CreateAccount/icons/SuccessSvg';
 import Badge, { BadgeProps } from '../Badge/Badge';
 
 interface InputContainerProps {
-  hasRight?: boolean;
-  isError?: boolean;
-  isSuccess?: boolean;
+  $hasRight?: boolean;
```
hcho112, line 10: is having dollar sign is necessary here?

Pessina: I was getting errors/warnings on the console telling me that hasRight, isSuccess... are being passed to the DOM and doesn't exist there. 

Then I checked the docs of styled components and looks like they also do it there: https://styled-components.com/docs/basics#adapting-based-on-props

But if you know a better way to do that we can try it too :)

hcho112: Based on the look, this component is being used on `/signin` endpoint right? I don't see the error/warning on console.log. Would you mind redirecting me where you found those errors/warnings?

Pessina: Screenshot 2024-01-06 at 7 37 07 PM

U are not getting this error on your console? If I use the `$` the console stops complaining

I'm using Chrome - Version 120.0.6099.199 (Official Build) (arm64)

hcho112: eh right, I was able to replicate it. I normally uses this repo along with near-discovery and warnings were happening within iframe.. 

In terms of actual problem, we actually had two approaches,
1. using full lower case, eg, `hasright` 
2. using $ in front eg. `$hasRight` 

Since styled component also recommends 2, I think it's LGTM.

hcho112 (approved):
LGTM

PR #146, opened 2024-01-06
Remove AddDevice screen

Pessina:
- Remove AddDevice screen and move it's logic to the Login screen. 
- Add Spinner for loading state
- Add forms validation on Login and CreateAccount screens
- Centralize send e-mails on VerifyEmail screen
- Clean up isRecovery search param from unnecessary places and use `userExist` and `recoverAccountWithOIDCToken` to check user existence


hcho112 (commented):
Great work! so code wise, it looks good. 
I have conducted following tests:
### Chrome Test
Create account
Sign in above
Sign in using cognito with above account
Sign in using safari browser with above account
Reach Delete key and continue sign in

### Safari Test
Sign in above

### Firefox test
Sign in
Create account

It looks good so far. For Firefox, we need [this PR](https://github.com/near/fast-auth-signer/pull/114) ready in order to check for FAK functionality check. 

I will leave @esaminu for final approval. (Please try to have a test independently on your side just in case)

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -10,90 +11,95 @@ import { redirectWithError, safeGetLocalStorage } from '../utils';
...
-    handleAuthState()
-      .catch((e) => {
+        setAuthenticated(await getAuthState(email, skipGetKeys));
+      } catch (e) {
+        captureException(e);
         redirectWithError({
           failure_url: failureUrl,
           success_url: successUrl,
           error:       e.message,
         });
```
esaminu, line 97: I know this was introduced in a separate PR but do we want to handle redirects from this hook? It strikes me as an unnecessary side effect that should be handled in the specific cases that require it outside this hook.

packages/near-fast-auth-signer/src/lib/useAuthState.ts
```diff
@@ -10,90 +11,95 @@ import { redirectWithError, safeGetLocalStorage } from '../utils';
...
-            networkId
-          });
-
-          await window.fastAuthController.setKey(new KeyPairEd25519(accountsList[0].keyPair.toString().split(':')[1]));
-          setAuthenticated(true);
-        } catch {
-          setAuthenticated(false);
+      try {
+        if (!email) {
+          throw new Error('Email is required');
```
esaminu, line 87: In the original implementation of this hook, we would return the auth state retrieved from the process of `getKeys` + indexer request etc in all cases except if the email in the query params is different from the locally stored `webauthn_username`. Why are we requiring the query param of the current page to have an email address here?

PR #147, opened 2024-01-08
Debounce API calls to validate email and username

Pessina:


packages/near-fast-auth-signer/package.json
```diff
@@ -50,6 +49,7 @@
     "https-browserify": "^1.0.0",
     "i18next": "^23.2.8",
     "js-sha256": "^0.10.1",
+    "lodash": "^4.17.21",
```
hcho112, line 52: instead of loading whole lib, lets just import debounce only
https://www.npmjs.com/package/lodash.debounce

Pessina: Good point, let me change it!

hcho112 (approved):
LGTM

PR #151, opened 2024-01-11
Update new testnet relayer url

hcho112:
Relayer URL for testnet has been updated a while ago but it hasn't been applied to this repo.

currently, test.near.org is pointing to old relayer api endpoint that no longer works.


itegulov (dismissed):

Pessina (approved):

hcho112:
Just to leave some comment, on testnet, we have been informed not to use SSL so it's `http` for now

PR #152, opened 2024-01-11
Check MPC Signature on claim OIDC token

Pessina:


hcho112:
Can we get one of you to review this verification step? @volovyks @itegulov ?
Cheers

packages/near-fast-auth-signer/src/utils/types.ts
```diff
@@ -22,6 +22,7 @@ type ProductionNetwork = {
     mpcRecoveryUrl: string;
     authHelperUrl: string; // TODO refactor: review by fastauth team
     accountIdSuffix: string;
+    mpcPublicKey: number[],
```
volovyks, line 25: Why not the PublicKyy type?

packages/near-fast-auth-signer/src/utils/mpc-service.ts
```diff
@@ -106,7 +107,14 @@ export const errorMessages: Record<string, string> = {
...
 
-export const MPC_PULIC_KEY = [
-  187, 24, 21, 82, 105, 165, 254, 26, 167, 89, 195, 236, 44, 83, 69, 87, 30,
-  151, 139, 229, 233, 182, 65, 230, 7, 234, 204, 91, 38, 70, 254, 254,
-];
+export const verifyMpcSignature = (mpcSignature: string, originalSignature: string): boolean => {
+  const mpcSignatureBytes = Uint8Array.from(Buffer.from(mpcSignature, 'hex'));
+  const originalSignatureBytes = Uint8Array.from(Buffer.from(originalSignature, 'hex'));
+  const claimData = { salt: CLAIM + 1, signature:  originalSignatureBytes };
+  const serializedData = serialize(new Map([
```
volovyks, line 114: I guess you are doing this conversion in other places, can we move it to a fn?

volovyks (approved):
@Pessina nice work!
The signature check part is good but may need light refactoring.
We do not need this right now, but in case we build a local testing environment, we will need to determine MPC PK on the flight.

packages/near-fast-auth-signer/src/utils/types.ts
Pessina: Good point, I will change it

packages/near-fast-auth-signer/src/utils/mpc-service.ts
Pessina: We are indeed serializing Maps in other places.

For the specific schema that we're using in `verifyMpcSignature`, it's unique and only used there. So, I think we should consider creating a reusable function when we encounter duplicated schema definitions.

As for the other schemas we're using, we might be able to create a reusable function for them as well. However, I need to investigate this further. Let's address this in a separate PR since those changes are unrelated to this feature.

What do you think?

volovyks (approved):

PR #153, opened 2024-01-12
Remove account exist check

hcho112:
This PR contains implementation for supporting `enable  email enumeration protection` on firebase. 

This change is part of security improvement where this change will stop from hackers to see if specific email is being used on fast-auth or not. 

From this change, regardless of email registered or not, it will trigger verify email check. If someone uses a specific email without create account first, it will still send email with verification link, but after link is being clicked, it will return with error:

image

In terms of post merge, we need to land this PR to production prior to enabling the configuration on firebase console.

### DIFF
--- a/packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
+++ b/packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
@@ -155,7 +155,6 @@ function CreateAccount() {
     const methodNames = searchParams.get('methodNames');
 
     try {
-      if (!isValid) return;
       const fullAccountId = `${data.username}.${network.fastAuth.accountIdSuffix}`;
       const {
         accountId
@@ -197,7 +196,7 @@ function CreateAccount() {
       //   title: message,
       // });
     }
-  }, [navigate, searchParams, isValid]);
+  }, [navigate, searchParams]);
 
   useEffect(() => {
     const email = searchParams.get('email');
--- a/packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
+++ b/packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
@@ -4,6 +4,7 @@ import { useCallback, useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { Link, useNavigate, useSearchParams } from 'react-router-dom';
 import styled from 'styled-components';
+import isEmail from 'validator/lib/isEmail';
 import * as yup from 'yup';
 
 import FormContainer from './styles/FormContainer';
@@ -13,7 +14,6 @@ import Input from '../../lib/Input/Input';
 import { openToast } from '../../lib/Toast';
 import { inIframe, redirectWithError } from '../../utils';
 import { network } from '../../utils/config';
-import { userExists } from '../../utils/firebase';
 import {
   accountAddressPatternNoSubAccount, getEmailId
 } from '../../utils/form-validation';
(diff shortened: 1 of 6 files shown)


packages/near-fast-auth-signer/src/components/Login/Login.tsx
```diff
@@ -50,17 +49,10 @@ function Login() {
     params: { email: string }
   ) => {
     try {
```
Pessina, line 51: As you removed the await, I believe we don't need the try/catch block anumore

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -4,6 +4,7 @@ import { useCallback, useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { Link, useNavigate, useSearchParams } from 'react-router-dom';
 import styled from 'styled-components';
+import isEmail from 'validator/lib/isEmail';
```
Pessina, line 7: We don't need to install an extra package, yup already haves a email validation. 

You can do 

```
yup
    .string()
    .required('Email address is required')
    .email('Invalid email')
```

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -160,6 +155,7 @@ function CreateAccount() {
     const methodNames = searchParams.get('methodNames');
 
     try {
+      if (!isValid) return;
```
Pessina, line 158: If we always call the createAccount with the handlerSubmit form hooks forms we are sure that it's a valid submission, so I believe we can remove this line.

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
hcho112: your suggestion doesn't block this type of emails:
`asodifnaiosdn@adiofnasoid.1234`

this PR's change does block:
image

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
hcho112: `createAccount` is also triggered by `useEffect`, which require condition to pause.

Pessina: In that case I think it's better to always call createAccount using handleSubmit(), so we can take advantage of the react-hooks-forms validation 

The way you implemented I believe the error won't be displayed to the user and we are duplicating code.

hcho112: I'm not sure if that's what we want. (I wasn't the one who implemented this part so I'm only guessing from previous implementation) 

I think having `useEffect` to run `createAccount` is actually good UX where, from previous screen, if email and account id is validated, user doesn't need to made fix and redirected to email verify page which is less of click. 

If we always force `handleSubmit` to do validation and click to go to next page, its another click of a button with extra screen. 

Personally, unless it is technically bad choice, I think current approach we have is good to move forward. (I always vote for better UX first)

Pessina: Oh no, we don't need to request the user to click the button.

It's just wrap the createAccount inside handleSubmit() inside the useEffect 

useEffect(() => {handleSub mit(createAccount)()}, [])

That way we make sure we run the validation before submitting. The email and username can come from the URL so the caller may not have run the validation before calling the fast-auth-signer

hcho112: agree! I have updated the changes and also tested that it work as expected

Pessina: Perfect, thanks Danny 😄

Pessina (approved):

Pessina (approved):

esaminu (approved):
LGTM

PR #156, opened 2024-01-22
fix: always validate url protocol

esaminu:
This PR validates urls per near-wallet implementation in https://github.com/near/near-wallet/blob/master/packages/frontend/src/utils/helper-api.js#L17


hcho112 (approved):
LGTM, good update!

PR #167, opened 2024-01-30
Firebase Account Indexer

Pessina:
- Introduce the Firebase Indexer as a Fallback for the Kitwallet (very unstable) for account retrieval using public keys.

### DIFF
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -18,10 +18,19 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
     throw new Error(`HTTP error! status: ${res.status}`);
   }
 
-  const accountIds = await res.json();
+  let accountIds: string[] = [];
+
+  // accountIds = await res.json();
+
+  if (accountIds.length === 0) {
+    const accountId = await window.firestoreController.getAccountIdByPublicKey(publicKey);
+    accountIds =  accountId ? [accountId] : [];
+  }
+
   if (accountIds.length === 0) {
     throw new Error('Unable to retrieve account id');
   }
+
   return accountIds;
 };
 
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -18,9 +18,7 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
     throw new Error(`HTTP error! status: ${res.status}`);
   }
 
-  let accountIds: string[] = [];
-
-  // accountIds = await res.json();
+  let accountIds: string[] = await res.json();
 
   if (accountIds.length === 0) {
     const accountId = await window.firestoreController.getAccountIdByPublicKey(publicKey);
--- a/packages/near-fast-auth-signer/src/lib/firestoreController.ts
+++ b/packages/near-fast-auth-signer/src/lib/firestoreController.ts
@@ -1,7 +1,7 @@
 import { captureException } from '@sentry/react';
 import { User } from 'firebase/auth';
 import {
-  getFirestore, Firestore, collection, setDoc, getDoc, getDocs, query, doc, CollectionReference, writeBatch
+  getFirestore, Firestore, collection, setDoc, getDoc, getDocs, query, doc, CollectionReference, writeBatch,
 } from 'firebase/firestore';
 import UAParser from 'ua-parser-js';
 
@@ -192,6 +192,23 @@ class FirestoreController {
   };
 
   getUserOidcToken = () => this.oidcToken;
+
+  async addPublicKey(publicKey: string, accountId: string) {
+    await setDoc(doc(this.firestore, 'publicKeys', publicKey), {
+      accountId,
+    });
+  }
+
+  async getAccountIdByPublicKey(publicKey: string): Promise<string> {
+    const publicKeyDoc = await getDoc(doc(this.firestore, 'publicKeys', publicKey));
+    if (!publicKeyDoc.exists()) {
+      return undefined;
+    }
+
+    const { accountId } = publicKeyDoc.data() as { accountId: string };
+
+    return accountId;
+  }
 }
 
 export default FirestoreController;
--- a/packages/near-fast-auth-signer/src/lib/firestoreController.ts
+++ b/packages/near-fast-auth-signer/src/lib/firestoreController.ts
@@ -1,7 +1,8 @@
 import { captureException } from '@sentry/react';
 import { User } from 'firebase/auth';
 import {
-  getFirestore, Firestore, collection, setDoc, getDoc, getDocs, query, doc, CollectionReference, writeBatch,
+  getFirestore, Firestore, collection, setDoc, getDoc, getDocs, query, doc, CollectionReference,
+  deleteDoc,
 } from 'firebase/firestore';
 import UAParser from 'ua-parser-js';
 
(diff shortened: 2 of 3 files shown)


packages/near-fast-auth-signer/src/lib/firestoreController.ts
```diff
@@ -47,12 +48,23 @@ class FirestoreController {
...
+    accountId?: string;
   }) {
     const parser = new UAParser();
     const device = parser.getDevice();
     const os = parser.getOS();
     const browser = parser.getBrowser();
     const dateTime = new Date().toISOString();
+
+    if (accountId && fakPublicKey) {
+      await window.firestoreController.addAccountIdPublicKey(fakPublicKey, accountId);
```
esaminu, line 65: Why not set this with the `device` and other metadata being set below? If it's for reusability in cases where we only set `accountId` we should chain the promise so that the same error handling applies

packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -18,10 +18,17 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
...
-  const accountIds = await res.json();
+  let accountIds: string[] = await res.json();
+
+  if (accountIds.length === 0) {
+    const accountId = await window.firestoreController.getAccountIdByPublicKey(publicKey);
+    accountIds =  accountId ? [accountId] : [];
+  }
+
   if (accountIds.length === 0) {
     throw new Error('Unable to retrieve account id');
```
hcho112, line 29: my assumption of using firebase is that if we use this, and ideally we will get 100% account Id return (for ones that exist), if it doesn't exist, instead of throw error, shouldn't we redirect them to verify email instead of throw error?

packages/near-fast-auth-signer/src/lib/firestoreController.ts
```diff
@@ -141,16 +153,16 @@ class FirestoreController {
...
         .filter((id) => id);
       if (firestoreIds.length) {
         // delete all records except the one that has LAK
-        firestoreIds.forEach((id) => {
-          batch.delete(doc(this.firestore, `/users/${this.userUid}/devices`, id));
-        });
-        await batch.commit();
+        const deletePromises = firestoreIds.flatMap((id) => [
+          deleteDoc(doc(this.firestore, `/users/${this.userUid}/devices`, id)),
+          deleteDoc(doc(this.firestore, '/publicKeys', id))
```
hcho112, line 163: have you tested that this code does delete the target public key correctly?

hcho112 (commented):
Also we need to think about existing users. Possibly we can either do one of three options:
1. If public key record found on firebase -> verify email again
2. Migrate existing public keys to firebase (not so practica tbh)
3. Maybe wait to build public key -> account id feature using QueryAPI?

I couldn't get it tested, but things that I want to confirm is:
1. Managing outdated/deleted public key. (As part of managing keys in fast-auth we delete keys when there are too many keys. (so we need to delete them when it gets deleted) 

But in general, say if QueryAPI does support a way to get public key -> account id feature, we would preferably want to use this instead. (Because this feature is being used on dApps that uses kitwallet and ideally we don't want to duplicate the same service.)

hcho112:
I manage to get the full end to end manual testing test working using https://github.com/near/near-discovery/pull/929 


Chrome:
:heavy_check_mark: Create account
:heavy_check_mark: Sign in on above account

Firefox (Non-webauthN):
:heavy_check_mark:  Check account 
:x: Sign in with an account created on Chrome
:x: Sign in with an account created on Firefox (itself)
(I think non-webauthN browser signin is broken, it hangs on `Signing transaction` text  view and stay there)

Safari:
:heavy_check_mark: Sign in with an account created on Chrome
:heavy_check_mark: Redirected to device page and go through full flow (Also confirmed that it does delete public key records well)
:x: Sign out and sign back in with same account (failed at kitwallet failure)

Result:
After conducting the full manual test, I have two things that need to point out for next step forward.
1. In this PR, we need to get non-webauthn flow fixed. Also ideally we need to replace every place in this repo that reference kit wallet for public key -> account id call. 
2. (Main problem) unfortunately making this PR/this repo fully replacing kitwallet won't fix the all problem. Because, the package `@near-js/biometric-ed25519``getKeys` method also uses kit wallet. But then, the point is that this package is design to be used with community (not tied to Pagoda internal usage) who doesn't support/use firebase. TDLR: if we really want to use firebase, we need to implement logic that try firebase if exist, or fall back to queryAPI(or kitwallet for now). CC @esaminu @DavidM-D

packages/near-fast-auth-signer/src/api/index.ts
Pessina: I believe this was introduce by you on the PR for non-WebauthN browsers (at least that's what my git says, please correct if I'm wrong). We can change it, but I would suggest doing it on another PR as this change it's not related to the current PR. 

Screenshot 2024-02-01 at 8 57 34 AM

Pessina:
@hcho112  I understand your concerns. But as I discussed with @esaminu, this PR isn't replacing KitWallet. We are introducing Firebase as a fallback in case the KitWallet can't return the accountId (due to recent hight instability).

As you can see on this [file](https://github.com/near/fast-auth-signer/blob/105320cf734da0746de2ee0b14c306749fa3fe1f/packages/near-fast-auth-signer/src/api/index.ts#L15), we first check if the account is available on KitWallet, if not, we check for Firebase. 

Firebase will only index accounts created after this PR, adding redundancy. Old accounts should be available by the KitWallet. 

The idea isn't index accounts created before this PR on Firebase.

packages/near-fast-auth-signer/src/lib/firestoreController.ts
Pessina: Good point, I will change it

hcho112:
@Pessina , [file](https://github.com/near/fast-auth-signer/blob/105320cf734da0746de2ee0b14c306749fa3fe1f/packages/near-fast-auth-signer/src/api/index.ts#L15) <- this implementation is exactly that I was talking about. But it looks like reference change isn't part of this PR or exist in `main` branch. I would suggest to above change to be merged to main soon or include in this PR together? (If so, this PR would be perfect!)

Maybe we shall discuss on the call together first

packages/near-fast-auth-signer/src/api/index.ts
hcho112: so I left this only because [file](https://github.com/near/fast-auth-signer/blob/105320cf734da0746de2ee0b14c306749fa3fe1f/packages/near-fast-auth-signer/src/api/index.ts#L15) wasn't part of this PR or wasn't expecting it to be coming. If we have this code, above can stay as it is

Pessina:
@hcho112 I didn't manage to reproduce the bugs you metioned, can you please re-test on your side and share a video or step-by-step how to reproduce the bugs? 

Setup: 

near-discovery: main branch
fast-auth-signer: 122-firebase-account-indexer branch

There are my testing videos

1 - Safari (Sign out and sign back in with same account (failed at kitwallet failure))
https://drive.google.com/file/d/1gdIcAZxz1mh6Py6QEa-XoYrStgd-YRkc/view?usp=sharing

2 - Firefox(Sign in with an account created on Firefox (itself))
https://drive.google.com/file/d/1BwWY9NVei22fvyQWkMj3eCwL9QlUU9bB/view?usp=sharing

3 - Chrome + Firefox (Sign in with an account created on Chrome)
https://drive.google.com/file/d/1mc64K23vQgFhZ2VKPCVl4ft9DcnWcr2F/view?usp=sharing

Please let me know if you don't have access to the recordings

packages/near-fast-auth-signer/src/lib/firestoreController.ts
Pessina: Updated :)

hcho112:
Thanks for the update and making records on each steps @Pessina 
First of all, I don't if it is from your latest change, but when I test again on FF, it worked finally. So I'm not against this PR going forward. 

Just feedback to your changes:
1. Interesting thing is that, when I tested it, I didn't even had to type the url manually, it actually redirected me straight back to where it was.
2. It works as expected
3. I can also see that it works. (Also unlike your video, it redirects within same browser tab so I didn't had to manually type anything) 

So at this point, I would say this is LGTM on my side. Lets follow up with @esaminu and ship it. Good work @Pessina

hcho112 (approved):
LGTM

esaminu (approved):
LGTM

PR #176, opened 2024-02-11
Initialize Firestore on App.tsx

Pessina:


hcho112:
I have pulled this PR locally and conducted a test with following condition:
1. Use current main branch on fast-auth-signer
2. Sign in with existing account on Chrome browser
3. Sign out 
4. Check out to current branch
5. Attempt to sign in

I see this error message:
image

Maybe we need to look into further?

hcho112:
above error is occurred from https://github.com/near/fast-auth-signer/blob/main/packages/near-fast-auth-signer/src/lib/controller.ts#L293

Maybe if user fail at this point, we send them email verification here? 
(eg, it would make sense to re-verify here because if expected oidc is absent, first thing we should do is email verification)

PR #178, opened 2024-02-20
Improve on account id retrival

hcho112:
This PR contains improvement on fetching account id from public key. 
1. Attempt to check with firebase first instead of kitwallet
2. Attempt to retrieve cached account id first before calling external source.

Background:
Currently the performance of the kitwallet is non-deterministic and we rather checks for firebase first and if there is no record, it will return response much quicker than kit wallet so overall it will improve UX.

### DIFF
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -10,25 +10,15 @@ import {
  *
  * @param publicKey - The public key to fetch the account IDs for.
  * @param options - An object containing the following properties:
- * - skipCache: A boolean indicating whether to skip the cache and fetch the account IDs from the external source.
  * - returnEmpty: A boolean indicating whether to return an empty array if no account IDs are found.
  * @returns A promise that resolves to an array of account IDs.
  * @throws Will throw an error if the fetch request fails.
  */
 
 type Option= {
-  skipCache?: boolean;
   returnEmpty?: boolean;
 }
 export const fetchAccountIds = async (publicKey: string, options?: Option): Promise<string[]> => {
-  if (window.fastAuthController && !options?.skipCache) {
-    const cachedAccountIds = await window.fastAuthController.getAccounts();
-
-    if (cachedAccountIds.length) {
-      return cachedAccountIds;
-    }
-  }
-
   // retrieve from firebase
   let accountIds = [];
   if (publicKey) {
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -9,25 +9,42 @@ import {
  * Fetches the account IDs associated with a given public key.
  *
  * @param publicKey - The public key to fetch the account IDs for.
+ * @param options - An object containing the following properties:
+ * - skipCache: A boolean indicating whether to skip the cache and fetch the account IDs from the external source.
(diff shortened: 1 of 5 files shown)


Pessina (dismissed):

hcho112:
@Pessina I have updated the PR based on your feedback. 
I have also made some enhancement on top of your feedback. On `fetchAccountId`, there is new arg called options (optional) where it allow `skipCache` and `returnEmpty`

skipCache -> force it to skip from retrieving locally, only use to get it from external source
returnEmpty -> force it to return empty accountIds instead of error it out

These are used in specific flow where:
When user authenticate/signin using passkey(finger print), based on its nature, it actually returns two keys and we need to figure out which one is the full access key that in on chain. This means that it will always have one key that is invalid. But we don't want to error this out and keep track error log because this is natural behaviour of web-authen based passkey.

packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -9,23 +9,42 @@ import {
...
-    throw new Error(`HTTP error! status: ${res.status}`);
-  }
 
-  let accountIds: string[] = await res.json();
+type Option= {
+  skipCache?: boolean;
+  returnEmpty?: boolean;
+}
+export const fetchAccountIds = async (publicKey: string, options?: Option): Promise<string[]> => {
+  if (window.firestoreController && !options?.skipCache) {
```
Pessina, line 24: Why do you check for the `window.firestoreController`? This isn't being used inside the if

I believe you should check for `window.fastAuthController`, no?

packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -9,23 +9,42 @@ import {
...
+  if (publicKey) {
     const accountId = await window.firestoreController.getAccountIdByPublicKey(publicKey);
-    accountIds =  accountId ? [accountId] : [];
+    if (accountId) {
+      return [accountId];
+    }
+  }
+
+  // retrieve from kitwallet
+  const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
```
Pessina, line 41: I would either 

`if (!publicKey) return []`

or 

Fetch from the KitWallet only if the publicKeys exists

packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -9,23 +9,42 @@ import {
...
-export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
-  const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
-  if (!res.ok) {
-    throw new Error(`HTTP error! status: ${res.status}`);
-  }
 
-  let accountIds: string[] = await res.json();
+type Option= {
+  skipCache?: boolean;
+  returnEmpty?: boolean;
```
Pessina, line 21: I like it. I think we should not throw an error on fetchAccountIds and this is a good start to allow us to refactor the codebase :) 

If we all agree on remove the throw error from here, we can clean up this option later

Pessina (changes requested):

packages/near-fast-auth-signer/src/api/index.ts
hcho112: oops, good catch!

esaminu (approved):
LGTM

PR #180, opened 2024-03-04
Initial Multi-chain implementation

hcho112:
DO NOT MERGE, this PR is currently created for review/discussion with the maintainers. 
This has been implemented along with:
https://gist.github.com/esaminu/8ecfd8524a3404d34135acde9c7cbb8b
https://gist.github.com/esaminu/f8cc37849de754f228c5a67bebce9b0f

Current out put
image

Todo:
1. Integrate with near-fastauth-wallet and use it against real world example
2. Update Infura account API
3. Any UI/UX improvements feedback
4. Check if gas fee calculation is correct. (It seems too low, but couldn't figure out what went wrong..)

Feel free to leave comments on things that are missing.


packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -40,11 +41,14 @@ const onCreateAccount = async ({
   email,
   gateway,
 }) => {
+  // Stop from LAK with multi-chain contract
+  const containBlacklistedContractId = getMultiChainContract() === contract_id;
```
esaminu, line 45: IMO we should throw earlier i.e. in the create account and add device pages to save the user the trouble. It should throw only if multichain contract is explicitly passed. If no value is passed we can continue the flow and only authenticate the signer

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const [deserializedDerivationPath, setDeserializedDerivationPath] = useState<DerivationPathDeserialized>(null);
+
+  // Send form height to modal if in iframe
+  useIframeDialogConfig({
+    element: signTransactionRef.current,
+    onClose: () => window.parent.postMessage({ signedDelegates: '', error:  'User cancelled action' }, '*')
+  });
+
+  const onError = (text: string) => {
+    window.parent.postMessage({ signedDelegates: '', text }, '*');
```
esaminu, line 51: this is a different payload than the error on line 47. Can we standardize and also would be useful if you documented the possible requests and responses in the PR or on a README on the folder

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -0,0 +1,226 @@
...
+      console.log('event', event);
+      // Maybe add origin check here? if url of origin is available, maybe update below
+      console.log('Message received in iframe: ', event.data);
+      if (event.data.data) {
+        setMessage(event.data.data);
+      }
+    };
+
+    window.addEventListener(
+      'multichain-message',
```
esaminu, line 67: should just be a `message` event

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const onError = (text: string) => {
+    window.parent.postMessage({ signedDelegates: '', text }, '*');
+    setError(text);
+  };
+
+  // event listener setup
+  useEffect(() => {
+    const handleMessage = (event) => {
+      console.log('event', event);
+      // Maybe add origin check here? if url of origin is available, maybe update below
```
esaminu, line 59: yes we should add a check for `event.origin` and compare with `domain` in the derivation path as part of the validation. If it's the same we can just proceed without prompting the user

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+  const oldPublicKeyPoint = ec.curve.point(x, y);
+
+  // Multiply the scalar by the generator point G
+  const scalarTimesG = ec.g.mul(scalarLittleEndian);
+
+  // Add the result to the old public key point
+  const newPublicKeyPoint = oldPublicKeyPoint.add(scalarTimesG);
+
+  return `04${
+    newPublicKeyPoint.getX().toString('hex')
```
esaminu, line 91: please update this with https://pagodaplatform.slack.com/archives/C04HWL4BJKE/p1709672855903079?thread_ts=1709638154.085769&cid=C04HWL4BJKE

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: I will try to make it similar withn SignMultichain component, but convention is already diverged on everywhere else in this repo. (eg. Existing Sign component)

hcho112: Created issue for handling above issue across whole repo https://github.com/near/fast-auth-signer/issues/183
Should be handled separately later

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: The reason why I gave different name is because on RpcRoute component handles add/remove event listener. Initially I thought I want to handle it separately, but actually its better to keep the same name, so will change this to 'message'

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: That check is actually handled in here: https://github.com/near/fast-auth-signer/pull/180/files#diff-d0f29bd93950dbf2e4531d2a1c32924daf12c3b34a972bd8b490bf032c283eabR139

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -146,8 +157,8 @@ function CreateAccount() {
     const success_url = searchParams.get('success_url');
     const failure_url = searchParams.get('failure_url');
     const public_key =  searchParams.get('public_key');
-    const contract_id = searchParams.get('contract_id');
     const methodNames = searchParams.get('methodNames');
+    const contract_id = searchParams.get('contract_id');
```
Pessina, line 161: This is duplicated, no? You are also defining it at line 123

packages/near-fast-auth-signer/src/components/CreateAccount/CreateAccount.tsx
```diff
@@ -119,6 +120,16 @@ function CreateAccount() {
   const [searchParams] = useSearchParams();
   const [inFlight, setInFlight] = useState(false);
 
+  const contractId = searchParams.get('contract_id');
+  useEffect(() => {
```
Pessina, line 124: This is duplicated, isn't better create a custom hook for it? 

Or call it only from the App.ts? (you set it to run only for routes that aren't `/multi-chain`)

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -106,8 +116,8 @@ function AddDevicePage() {
     const success_url = searchParams.get('success_url');
     const failure_url = searchParams.get('failure_url');
     const public_key =  searchParams.get('public_key');
-    const contract_id = searchParams.get('contract_id');
     const methodNames = searchParams.get('methodNames');
+    const contract_id = searchParams.get('contract_id');
```
Pessina, line 120: Duplicated, declared above too, no?

packages/near-fast-auth-signer/src/components/SignMultichain/MultichainExecutor.ts
```diff
@@ -0,0 +1,33 @@
+import { EVMInterface, BTCInterface, MultichainExecutionResponse, InitializationConfig } from './types';
```
Pessina, line 1: I think you are already planning it, but just to not. This can be removed

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/SignMultichain/schema.ts
```diff
@@ -0,0 +1,6 @@
+import { BorshSchema } from 'borsher';
+
+export const derivationPathSchema = BorshSchema.Struct({
```
Pessina, line 3: I think this is important to be on the docs, so the caller knows the schema.

packages/near-fast-auth-signer/src/components/SignMultichain/types.ts
```diff
@@ -0,0 +1,57 @@
+import { KeyPair } from 'near-api-js';
+
+export interface DerivationPathDeserialized {
+  asset: 'ETH' | 'BNB' | 'BTC';
+  domain?: string;
+}
+
+export interface MultichainExecutionResponse {
```
Pessina, line 8: I think some of those types you can clean up. Like this is only used here. Please also check the other types

packages/near-fast-auth-signer/src/components/SignMultichain/bitcoin/schema.ts
```diff
@@ -0,0 +1,17 @@
+import * as yup from 'yup';
+
+export const bitcoinSchema = yup.object().shape({
+  derivationPath:       yup.string().required('derivationPath is required'),
+  to:                   yup.string().required('to is required'),
+  value:                yup.mixed<bigint>().required('value is required'),
+  derivedPublicKey:     yup.string().required('derivedPublicKey is required'),
```
Pessina, line 7: I think this is not necessary anymore, you can derive it from the derivationPath

packages/near-fast-auth-signer/src/components/SignMultichain/schema.ts
```diff
@@ -0,0 +1,6 @@
...
+  asset:         BorshSchema.String,
```
Pessina, line 4: I'm not sure if it should be a string. 

I think it's better create a enum of supported chains and re-use it on all the places where you do chain checks.

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -0,0 +1,193 @@
...
+  return Number(value);
+};
+
+export const getTokenAndTotalPrice = async (asset: DerivationPathDeserialized['asset'], value: MultichainInterface['value']) => {
+  const id = multichainAssetToCoinGeckoId(asset);
+  if (id) {
+    const res = await getMultichainCoinGeckoPrice(asset);
+    if (!res) {
+      return {
+        price:       0,
```
Pessina, line 101: If you can't get the data, I think it's better to use null/undefined so it's clear for the caller that the data couldn't be fetched. 0 still a number

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -0,0 +1,193 @@
...
+    BNB:  'Binance Smart Chain',
+    BTC:  'Bitcoin Network',
+  };
+
+  return map[asset] || null;
+};
+
+export const getMultichainCoinGeckoPrice = async (asset: DerivationPathDeserialized['asset']) => fetchGeckoPrices(multichainAssetToCoinGeckoId(asset));
+
+const convertTokenToReadable = (value : MultichainInterface['value'], asset: DerivationPathDeserialized['asset']) => {
```
Pessina, line 85: Isn't better `getTokenAmount()` I think it's more readable and down there you do

`const tokenAmount = convertTokenToReadable()`

I think it sounds more readable

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -0,0 +1,193 @@
...
+    const tokenPrice: number = res[id].usd;
+    const tokenAmount = convertTokenToReadable(value, asset);
+    return {
+      price: parseInt((tokenPrice * tokenAmount * 100).toString(), 10) / 100,
+      tokenAmount,
+      tokenPrice
+    };
+  }
+  return {
+    price:       0,
```
Pessina, line 114: Same as above

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -0,0 +1,193 @@
...
+    tokenAmount: convertTokenToReadable(value, asset)
+  };
+};
+
+export const shortenAddress = (address: string): string => {
```
Pessina, line 119: Screenshot 2024-03-16 at 1 23 50 AM

There are some duplicated numbers, is this a bug? I think the user would expect the two parts of the string to be not overlapping. 

That's what I would expect

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/SignMultichain/schema.ts
Pessina: It's ok if you don't do on this PR, but create a issue for this and we do the docs later.

Pessina (dismissed):

Pessina (approved):

PR #181, opened 2024-03-06
Improve account id fetch with race condition

hcho112:
This PR contains implementation for improve UX of fetching account id. 

As part of using web-authN passkey, we always given two sets of key-pairs and only one of them are genuine. We find the true one by checking its public key presence on firebase (main) AND kitwallet (fallback). The issue was that on previous implementation, we use `Promise.allSettled` to wait for both keys to be finished with checking with firebase + kitwallet. But the problem here is that, one of the key is falsy, in which it does not exist in firebase store, and therefore always hit kitwallet. As of now, kitwallet is the bottleneck of UX where it often takes VERY long to process or sometimes returns invalid result. Its performance is non-deterministic and ideally we should avoid depending on it as much as possible for now. 

Solution on this PR is by having `Promise.race` to grap what gets returned first. Right now, out of given two keys, time taken to get genuine key is ALWAYS faster so once we receive it, we check for response with presence of account Id. If present, we return immediately, otherwise we also checks for second key just in case (but only if first one returns with empty)

Also there is an improvement on `fetchAccountIds` where it also uses `Promise.race` with wait duration of 10 seconds on kit wallet to response. If it doesn't return within 10 seconds, it will consider it as falsy and return empty. (If so, user will be redirected to email verification flow)

These changes will resolve one of two major bottleneck of fast-auth experience.

### DIFF
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -5,15 +5,8 @@ import { network } from '../utils/config';
 import {
   CLAIM, getUserCredentialsFrpSignature
 } from '../utils/mpc-service';
+import { withTimeout } from '../utils';
 
-// Use this function to implement wait logic for async process
-const withTimeout = async (promise, timeoutMs) => {
-  // Create a promise that resolves with false after timeoutMs milliseconds
-  const timeoutPromise = new Promise((resolve) => { setTimeout(() => resolve(false), timeoutMs); });
-
-  // Race the input promise against the timeout
-  return Promise.race([promise, timeoutPromise]);
-};
 
 /**
  * Fetches the account IDs associated with a given public key.
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -1,3 +1,4 @@
+import { captureException } from '@sentry/react';
 import { KeyPair } from 'near-api-js';
 
 import { network } from '../utils/config';
@@ -31,14 +32,18 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
       }
     }
 
-    return withTimeout(await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION)
-      .then(async (res) => {
-        if (res) {
-          const ids = await res.json();
-          return ids;
-        }
-        return [];
-      });
+    try {
+      const res = await withTimeout(fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION);
+      if (res) {
+        const ids = await res.json();
+        return ids;
+      }
+      return [];
+    } catch (error) {
+      console.log('fetchAccountIds', error);
+      captureException(error);
+      return [];
+    }
   }
 
   return [];
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -1,12 +1,11 @@
 import { captureException } from '@sentry/react';
 import { KeyPair } from 'near-api-js';
 
+import { withTimeout } from '../utils';
 import { network } from '../utils/config';
 import {
   CLAIM, getUserCredentialsFrpSignature
 } from '../utils/mpc-service';
-import { withTimeout } from '../utils';
-
 
 /**
  * Fetches the account IDs associated with a given public key.
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -5,22 +5,24 @@ import {
   CLAIM, getUserCredentialsFrpSignature
 } from '../utils/mpc-service';
 
+// Use this function to implement wait logic for async process
+const withTimeout = async (promise, timeoutMs) => {
+  // Create a promise that resolves with false after timeoutMs milliseconds
+  const timeoutPromise = new Promise((resolve) => { setTimeout(() => resolve(false), timeoutMs); });
+
+  // Race the input promise against the timeout
+  return Promise.race([promise, timeoutPromise]);
+};
+
 /**
  * Fetches the account IDs associated with a given public key.
  *
  * @param publicKey - The public key to fetch the account IDs for.
- * @param options - An object containing the following properties:
- * - returnEmpty: A boolean indicating whether to return an empty array if no account IDs are found.
  * @returns A promise that resolves to an array of account IDs.
  * @throws Will throw an error if the fetch request fails.
  */
-
-type Option= {
-  returnEmpty?: boolean;
-}
-export const fetchAccountIds = async (publicKey: string, options?: Option): Promise<string[]> => {
-  // retrieve from firebase
-  let accountIds = [];
+const KIT_WALLET_WAIT_DURATION = 10000; // 10s
+export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
   if (publicKey) {
     if (window.firestoreController) {
       const accountId = await window.firestoreController.getAccountIdByPublicKey(publicKey);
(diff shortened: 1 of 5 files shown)


packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -5,42 +5,82 @@ import {
...
+/**
+ * This function fetches account id from given two public keys.
+ * Webuathn currently prompts two public keys by default and we need to discover which key is currently on chain.
+ * Since we store public key <-> account id in firestore, one of the public key will return the result fast and we do not need to wait for both request to resolve.
+ * It is possible that both public keys are not on chain, in that case we return null.
+ * @param publicKeyA
+ * @param publicKeyB
+ * @returns object with account id and public key
+ */
+export const fetchAccountIdsFromTwoKeys = async (
```
Pessina, line 56: I think this is a very complicated code and case specific, why don't we let the caller manage it? 

We can do something like: 

```
function timeoutPromise(duration: number): Promise<never> {
  return new Promise((_, reject) => {
    setTimeout(() => reject(new Error('Timeout')), duration);
  });
}

 const accountPromises = keypairs.map((keypair) => Promise.race([
        fetchAccountIds(keypair.getPublicKey().toString()),
        timeoutPromise(10000)
      ]).catch(() => null));

      const accountResults = await Promise.all(accountPromises);
      const accountInfo = accountResults.reduce((acc, result, index) => {
        if (result && result.length > 0) {
          return {
            accId:     result[0],
            publicKey: keypairs[index].toString(),
          };
        }
        return acc;
      }, null);
```

This is just an example, the Promise.race with timeout should stay inside the fetchAccountIds function as you did in your code.

packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -5,42 +5,82 @@ import {
...
       }
     }
-    // retrieve from kitwallet
-    const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
-    if (!res.ok) {
-      throw new Error(`HTTP error! status: ${res.status}`);
-    }
-    accountIds = await res.json();
+
+    return withTimeout(await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION)
```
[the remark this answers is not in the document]
hcho112: good point, will convert this to async await fashion

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -230,10 +229,14 @@ function AuthCallbackPage() {
           setStatusMessage(isRecovery ? 'Recovering account...' : 'Creating account...');
 
           // claim the oidc token
-          window.fastAuthController = new FastAuthController({
-            accountId,
-            networkId
-          });
+          if (window.fastAuthController) {
```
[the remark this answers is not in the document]
hcho112: This was to prevent from re-declaring the `FastAuthController` class. If we re-declare, existing in memory key pair gets flushed out. Instead, just update account id if logic reaches upto this point.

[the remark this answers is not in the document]
Pessina: Got it, can we create something re-usable for this? This code it's being repeated 3 times on your PR

packages/near-fast-auth-signer/src/api/index.ts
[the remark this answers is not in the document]
Pessina: Cool, thanks. ;)

packages/near-fast-auth-signer/src/api/index.ts
hcho112: Based on my understanding, above example will mostly take 10s (with current status of kitwallet) because of `Promise.all(accountPromises);`. Compare that against my solution, it will return as soon as it finds the account id from firebase. 

I don't think it is complicated. You could say it is case specific as it is being used on single place, but it is just simply abstraction of optimisation of account id retrieval. I still think it is still valuable to keep this as it is and make the `useAuthState.ts` simple and readable as possible. (currently `useAuthState.ts` is fairly logic heavy as it is)

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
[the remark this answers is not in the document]
hcho112: updated the syntax

packages/near-fast-auth-signer/src/hooks/useAuthState.ts
```diff
@@ -28,26 +28,25 @@ export const getAuthState = async (email?: string | null): Promise<AuthState> =>
...
-      if (accountsList.length === 0) {
+      if (!accountInfo.accId) {
         return false;
       }
 
-      window.fastAuthController = new FastAuthController({
-        accountId: accountsList[0].accId,
-        networkId
-      });
+      if (!window.fastAuthController) {
```
Pessina, line 40: I think you forgot to re-use the function here too

packages/near-fast-auth-signer/src/api/index.ts
Pessina: I'm sorry I wasn't clear when I said case specific, my bad. 

When I said case specific it's that we can only use this for the publicKey and only 2 keys. 

I believe this should be re-usable on the sense of you can pass as many API calls as you want and race all of them, making this function more flexible and re-usable. 

It should be something like 

`function race<T>(promisesArr: Promise<T>[], timeout: number): Promise<{index: number, value: T}> {}`

Then on this case we call it with: 

`const { accountId, index }  = await race([fetchAccountIds(publicKeyA), fetchAccountIds(publicKeyB)], 10000)`

I'm not 100% if my types are correct, but if it's not clear please let me know

--------------------

For the part of not making `useAuthState.ts` more complicated, I fully agree. I think we should idd abstract this logic out of it, it's a good point!

Pessina: Also, 

I think this function and `withTimout` should not live on this file `api/index.ts`. This file should be dedicated only to API calls not utils functions.

Pessina (changes requested):

packages/near-fast-auth-signer/src/api/index.ts
Pessina: I'm sorry I wasn't clear when I said case specific, my bad.

When I said case specific it's that we can only use this for the publicKey and only 2 keys.

I believe this should be re-usable on the sense of you can pass as many API calls as you want and race all of them, making this function more flexible and re-usable.

It should be something like

function race<T>(promisesArr: Promise<T>[], timeout: number): Promise<{index: number, value: T}> {}

Then on this case we call it with:

const { accountId, index }  = await race([fetchAccountIds(publicKeyA), fetchAccountIds(publicKeyB)], 10000)

I'm not 100% if my types are correct, but if it's not clear please let me know

---------------------------------------------

For the part of not making useAuthState.ts more complicated, I fully agree. I think we should idd abstract this logic out of it, it's a good point!

Pessina: Also I think this code and the `withTimout` should not live on the `api/index.ts` file as those are utils funcitons and this file should be dedicated to API calls.

hcho112: - I believe this should be re-usable on the sense of you can pass as many API calls as you want and race all of them, making this function more flexible and re-usable.
- 
- It should be something like
- 
- function race<T>(promisesArr: Promise<T>[], timeout: number): Promise<{index: number, value: T}> {}
- 
- Then on this case we call it with:
- 
- const { accountId, index }  = await race([fetchAccountIds(publicKeyA), fetchAccountIds(publicKeyB)], 10000)
- 
- I'm not 100% if my types are correct, but if it's not clear please let me know

@Pessina thanks for the feedback, but I disagree with above approach with few reasons:
1. This function is implemented for handling specific edge case for account Id retrieval from two keys (Please let me know if you aren't sure why we are always getting two keys).
2. This function is not just simply returning firstly executed async function but also does check for returned value and if it is empty, it does check for following async function.
3. `withTimeout` function already fulfil the re-usable utility perspective of race condition & timeout feature of async function.

Pessina: I understand what this function is doing. 

My point is: imagine a new scenario, in the future, where we also have to race 3 API calls of the MPC Recovery. Will we create another custom function for this new scenario? 

Creating a re-usable function that accepts an array of Promises allow us to re-use it anytime we need to race Promises, without any `amount of promises` or `type` restriction.

Pessina: @hcho112 this is not really a blocker to me as it's isolated. But I see this a nice improvement that can be done in a few minutes and create a nice pattern for the future. 

It would be really bad if we keep creating one function per Promise type that we have to race.

Pessina (dismissed):

Pessina (approved):

PR #186, opened 2024-03-09
feat: remove the signer as a dependency

esaminu:
- [x] Make input/output/fee/gasLimit  optional and provided by the caller {expose BTC fee}
- [x] Move signDelegate code out of Felipe code to create npm package. `signAndSend` should take just a `KeyPair`
- [x] Extract KDF to separated file and export function individually
- [x] Use mpc contract address and fetch root key using the provided address
- [x] Breakup `signAndSend` into smaller functions
- [x] Remove `Account` instance and `relayerUrl` from being required in `getDerivedAddress`
- [x] Remove `relayerUrl` from being required in `getEstimatedFeeEVM`

### DIFF
--- a/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
+++ b/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
@@ -12,7 +12,10 @@ import { assertNever } from '../../utils';
 import { networkId } from '../../utils/config';
 import { Bitcoin } from '../../utils/multi-chain/chains/Bitcoin/Bitcoin';
 import {
-  getBitcoinFeeProperties, getEstimatedFeeEVM, signAndSendBTCTransaction, signAndSendEVMTransaction
+  fetchBTCFeeProperties,
+  fetchEstimatedEVMFee,
+  signAndSendBTCTransaction,
+  signAndSendEVMTransaction
 } from '../../utils/multi-chain/multiChain';
 import { fetchGeckoPrices } from '../Sign/Values/fiatValueManager';
 
@@ -194,14 +197,14 @@ export const multichainGetTotalGas = async ({
   from?: string;
 }) => {
   if (asset === 'BTC') {
-    const satoshis =  (await getBitcoinFeeProperties(CHAIN_CONFIG.BTC.providerUrl, from, [{
+    const satoshis =  (await fetchBTCFeeProperties(CHAIN_CONFIG.BTC.providerUrl, from, [{
       address: to,
       value:   Number(value)
     }])).fee;
 
     return Bitcoin.toBTC(satoshis);
   } if (isEVMChain(asset)) {
-    const wei = await getEstimatedFeeEVM({
+    const wei = await fetchEstimatedEVMFee({
       to,
       value
     }, {
--- a/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
+++ b/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
@@ -10,18 +10,21 @@ import {
 } from './types';
 import { assertNever } from '../../utils';
 import { networkId } from '../../utils/config';
-import { Bitcoin } from '../../utils/multi-chain/chains/Bitcoin/Bitcoin';
 import {
   fetchBTCFeeProperties,
+  fetchEVMFeeProperties,
   signAndSendBTCTransaction,
-  signAndSendEVMTransaction
+  signAndSendEVMTransaction,
 } from '../../utils/multi-chain/multiChain';
-import { fetchEVMFeeProperties } from '../../utils/multi-chain/utils';
 import { fetchGeckoPrices } from '../Sign/Values/fiatValueManager';
 
 // TODO: use this for blacklisting on limited access key creation AND sign
 const MULTICHAIN_CONTRACT_TESTNET = 'multichain-testnet-2.testnet';
 const MULTICHAIN_CONTRACT_MAINNET = 'multichain-testnet-2.testnet';
+
+function toBTC(satoshis: number): number {
+  return satoshis / 100000000;
+}
 
 type BTCFeeProperites = {
   inputs: {
@@ -127,7 +130,7 @@ const convertTokenToReadable = (value : MultichainInterface['value'], asset: Cha
     return parseFloat(formatEther(value));
   }
   if (asset === 'BTC') {
-    return Bitcoin.toBTC(Number(value));
+    return toBTC(Number(value));
   }
   return Number(value);
 };
@@ -232,7 +235,7 @@ export const multichainGetFeeProperties = async ({
       value:   Number(value)
     }]));
 
-    return { ...feeProperties, feeDisplay: Bitcoin.toBTC(feeProperties.fee) };
+    return { ...feeProperties, feeDisplay: toBTC(feeProperties.fee) };
   } if (isEVMChain(asset)) {
     const feeProperties = await fetchEVMFeeProperties(CHAIN_CONFIG.ETH.providerUrl, {
       to,
--- a/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
+++ b/packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
@@ -13,15 +13,38 @@ import { networkId } from '../../utils/config';
 import { Bitcoin } from '../../utils/multi-chain/chains/Bitcoin/Bitcoin';
 import {
   fetchBTCFeeProperties,
-  fetchEstimatedEVMFee,
   signAndSendBTCTransaction,
   signAndSendEVMTransaction
 } from '../../utils/multi-chain/multiChain';
+import { fetchEVMFeeProperties } from '../../utils/multi-chain/utils';
 import { fetchGeckoPrices } from '../Sign/Values/fiatValueManager';
 
 // TODO: use this for blacklisting on limited access key creation AND sign
 const MULTICHAIN_CONTRACT_TESTNET = 'multichain-testnet-2.testnet';
 const MULTICHAIN_CONTRACT_MAINNET = 'multichain-testnet-2.testnet';
+
+type BTCFeeProperites = {
+  inputs: {
+      txid: string;
+      vout: number;
+      value: number;
+      script: string;
+  }[];
+  outputs: {
+      address: string;
+      value: number;
+  }[];
+  fee: number;
+};
+
+type EVMFeeProperties = {
+  gasLimit: bigint;
+  maxFeePerGas: bigint;
+  maxPriorityFeePerGas: bigint;
+  maxFee: bigint;
+}
+
+export type TransactionFeeProperties = BTCFeeProperites | EVMFeeProperties
 
 const EVMChains: EVMChainMap<boolean> = {
   ETH: true,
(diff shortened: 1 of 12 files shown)


packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -30,14 +33,16 @@ export const sign = async (
     new BN(0)
   );
 
-  const signedDelegate = await window.fastAuthController.signDelegateAction(
+  const signedDelegate = await account.signedDelegate(
     {
-      receiverId: 'multichain-testnet-2.testnet',
-      actions:    [functionCall],
-      signerId:   account.accountId
+      receiverId:     'multichain-testnet-2.testnet',
```
esaminu, line 38: Should be received from config

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
```diff
@@ -123,12 +124,13 @@ export class Bitcoin {
...
    */
-  async fetchUTXOs(
+  static async fetchUTXOs(
+    providerUrl: string,
     address: string
   ): Promise<UTXO[]> {
     try {
       const response = await axios.get(
-        `${this.providerUrl}address/${address}/utxo`
+        `${providerUrl}address/${address}/utxo`
```
esaminu, line 133: We need to state explicitly that the standard for this module is trailing slash

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -1,57 +1,11 @@
...
-
-type FailureResponse = {
-  success: false;
-  errorMessage: string;
-}
-
-type Response = SuccessResponse | FailureResponse
-
-const MPC_ROOT_PUBLIC_KEY =  'secp256k1:4HFcTSodRLVCGNVcGc4Mf2fwBBBxv9jxkGdiW2S2CA1y6UpVVRWKj6RX7d7TDt65k2Bj3w9FU4BGtt43ZvuhCnNt';
+const MPC_ROOT_PUBLIC_KEY = 'secp256k1:4HFcTSodRLVCGNVcGc4Mf2fwBBBxv9jxkGdiW2S2CA1y6UpVVRWKj6RX7d7TDt65k2Bj3w9FU4BGtt43ZvuhCnNt';
```
esaminu, line 8: the mpc contract accountId needs to be passed in as a config and we need to fetch this from there

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -1,57 +1,10 @@
...
 
 const signAndSend = async (req: Request): Promise<Response> => {
```
esaminu, line 9: As discussed we should break this up into multiple function so we don't create an implicit standard with the `type` key to set the asset

esaminu:
As discussed we should also break up the kdf in `getDerivedAddress` into per chain functions

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -1,82 +1,59 @@
...
-
-      txid = (await evm.handleTransaction(
-        req.transaction,
-        req.account,
-        req.transaction.derivedPath,
-      )).hash;
-    } else if (req.chainConfig.type === 'BTC') {
-      const btc = new Bitcoin({ ...req.chainConfig, relayerUrl: req.fastAuthRelayerUrl });
+    const btc = new Bitcoin({
+      ...(req.chainConfig as BTCChainConfigWithProviders),
```
Pessina, line 48: Same here, we should not use casting inside chain specific methods. The Request args should be BTC specific

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -85,41 +62,30 @@ const signAndSend = async (req: Request): Promise<Response> => {
...
-      )).address;
-      break;
-    default:
-      throw new Error('Unsupported chain config');
-  }
-
-  return derivedAddress;
-};
+  derivationPath: string,
+  bitcoinNetwork: 'mainnet' | 'testnet',
```
Pessina, line 73: Better extract this to a type. BTCNetwork same for NEAR

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -85,41 +62,30 @@ const signAndSend = async (req: Request): Promise<Response> => {
...
+    derivationPath,
+    bitcoinNetwork === 'testnet'
+      ? bitcoin.networks.testnet
+      : bitcoin.networks.bitcoin,
+    nearNetworkId,
+    multichainContractId
+  )
+).address;
+
+export const getDerivedEVMAddress = EVM.deriveAddress;
```
Pessina, line 88: For this static methods, isn't better simply expose the class? Doing what we are doing here it's just adding more code to `maintain`, like the getDerivedBTCAddress. 

If the function or static method interface changes we will have to update it in two places. I think it's better simply expose the class method

We can have a chat on it. Because I'm not 100% happy with making most of the class method static, then we lose the class purpose.

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -1,12 +1,15 @@
+import { Account, Connection, Contract } from '@near-js/accounts';
```
Pessina, line 1: Amazing work on this file!! I thought it would be much more complicated.

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -48,14 +51,39 @@ export const sign = async (
...
-      receiverId: contract,
-      actions:    [functionCall],
-      signerId:   account.accountId
+      receiverId:     contract,
+      actions:        [functionCall],
+      blockHeightTtl: 60,
     }
   );
 
+  // TODO: add support for creating the signed delegate using the mpc recovery service with an oidc_token
```
Pessina, line 85: Do we need to fix this before merge?

Pessina (changes requested):

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
```diff
@@ -1,10 +1,10 @@
 import axios from 'axios';
```
Pessina, line 1: I believe there is a type issue


https://github.com/near/fast-auth-signer/assets/22358004/a26b508e-89a0-49b1-8ad3-231dc6bbb417

I think when you call `attatchGasAndNonce` you should parse the value to ethers `parseEthers(data.value)`

Pessina (changes requested):

packages/near-fast-auth-signer/src/utils/multi-chain/chains/EVM.ts
```diff
@@ -106,17 +109,19 @@ class EVM {
    * @returns {Promise<ethers.providers.TransactionRequest>} A new transaction object augmented with gas price, gas limit, chain ID, and nonce.
    */
   async attachGasAndNonce(
-    transaction: ethers.TransactionLike & {
-      from: string;
-    }
+    transaction: EVMTransaction & { from: string }
```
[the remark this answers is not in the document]
esaminu: The difference is that in our input we allow the user to provide us with a nice eth string and convert to Wei bigint for them instead of requiring them to pass that

packages/near-fast-auth-signer/src/utils/multi-chain/chains/types.ts
```diff
@@ -0,0 +1,68 @@
...
+  inputs: UTXO[];
+  outputs: { address: string, value: number }[];
+};
+
+export type BTCTransaction = BaseTransaction & (BtcInputsAndOutputs | {
+  inputs?: never;
+  outputs?: never;
+});
+
+type ChainProviders = {
```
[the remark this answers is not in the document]
esaminu: We’ll add this in next iteration because currently we still don’t support not having a passkey FAK in the browser and trying to use sign or sign multi chain in the signer app

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin/Bitcoin.ts
```diff
@@ -428,12 +375,11 @@ export class Bitcoin {
       },
     };
 
-    // TODO: it should be done in parallel,
-    // but for now it's causing nonce issues on the signDelegate so we will run sequentially to avoid the issue for now
-    for (let index = 0; index < inputs.length; index += 1) {
-      // eslint-disable-next-line no-await-in-loop
-      await psbt.signInputAsync(index, mpcKeyPair);
-    }
+    await Promise.all(
```
Pessina, line 378: I think there is a rebase issue here, we can't send the requests in parallel because of nonce issues. I talked with @hcho112 and we will create a separated issue for it

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin/Bitcoin.ts
```diff
@@ -343,33 +293,30 @@ export class Bitcoin {
...
-      value: number;
-    },
-    account: Account,
+    data: BTCTransaction,
+    nearAuthentication: NearAuthentication,
     path: string,
   ) {
-    const { address, publicKey } = await Bitcoin.deriveAddress(
-      account.accountId,
+    const satoshis = Bitcoin.toSatoshi(parseFloat(data.value));
```
Pessina, line 307: The value now it's passed as satoshis, we can remove this line. also rebase issue I think. Update the docs too

packages/near-fast-auth-signer/src/utils/multi-chain/chains/EVM/EVM.ts
```diff
@@ -106,14 +96,29 @@ class EVM {
...
-    transaction: ethers.TransactionLike & {
-      from: string;
-    }
+    transaction: EVMTransaction & { from: string }
   ): Promise<ethers.TransactionLike> {
-    const { gasLimit, maxFeePerGas, maxPriorityFeePerGas } = await this.getFeeProperties(transaction);
-    const nonce = await this.provider.getTransactionCount(transaction.from, 'latest');
+    const parsedValueTransaction = {
+      ...transaction,
+      value: parseEther(transaction.value),
```
Pessina, line 103: This is already coming in ethers, we don't need to parseAnymore

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
```diff
@@ -10,14 +9,43 @@ import {
...
+} from '../../utils/multi-chain/multiChain';
+import { fetchEVMFeeProperties } from '../../utils/multi-chain/utils';
 import { fetchGeckoPrices } from '../Sign/Values/fiatValueManager';
 
 // TODO: use this for blacklisting on limited access key creation AND sign
 const MULTICHAIN_CONTRACT_TESTNET = 'multichain-testnet-2.testnet';
 const MULTICHAIN_CONTRACT_MAINNET = 'multichain-testnet-2.testnet';
 
+type BTCFeeProperites = {
+  inputs: {
```
Pessina, line 27: Re-use UTXO type here

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/SignMultichain/utils.ts
esaminu: Don't think we should export anything from the module aside from the multi chain file for now to prevent unpredictable usage paths

Pessina (approved):

PR #187, opened 2024-03-12
Add Support for Multi-Chain Transactions in near-fast-auth-signer

Pessina:
This PR adds multi-chain transaction support to near-fast-auth-signer, enabling users to sign and send transactions on Ethereum, BSC, and Bitcoin testnet using NEAR accounts.

Changes include:

- New utility functions in multiChain.ts for handling chain signature
- Chain-specific modules for EVM chains and Bitcoin in chains directory
- kdf.ts for deriving Ethereum and Bitcoin addresses from root key
- signAndSend function to simplify signing and broadcasting transactions

Developers can now easily integrate multi-chain capabilities into their apps, allowing interaction with Ethereum, BSC, and Bitcoin testnet using NEAR accounts.


packages/near-fast-auth-signer/src/components/Test/test.tsx
hcho112, line 1: should we delete this file and manually use test code on our local machine?

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
```diff
@@ -0,0 +1,448 @@
...
+  ): Promise<{
+    inputs: UTXO[],
+    outputs: {address: string, value: number}[],
+    fee: number,
+  }> {
+    const utxos = await this.fetchUTXOs(from);
+    const feeRate = await this.fetchFeeRate(confirmationTarget);
+
+    // TODO: test net won't work with feeRate fetched, it need to add 1
+    const ret = coinselect(utxos, targets, feeRate + 1);
```
hcho112, line 331: if + 1 is only todo with testnet, can we implement a logic to check if `config.networkType === 'testnet' ? feeRate + 1 : feeRate`

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
```diff
@@ -0,0 +1,448 @@
...
+    });
+
+    data.vin.forEach((vin, index) => {
+      if (vin.witness && vin.witness.length > 0) {
+        const witness = vin.witness.map((w) => Buffer.from(w, 'hex'));
+        tx.setWitness(index, witness);
+      }
+    });
+
+    return tx;
```
hcho112, line 206: so in fast-auth-signer app, if I want hashed transaction, do I use this function and then encode with base64?

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Calculates the estimated fee for a Bitcoin transaction in satoshis.
+ *
+ * @param {Object} transaction - The transaction details including the sender's address and the targets.
+ * @param {string} transaction.from - The Bitcoin address from which the transaction is sent.
+ * @param {Array} transaction.targets - An array of objects, each containing the target address and value in satoshis to send.
+ * @param {BTCChainConfigWithProviders} chainConfig - The configuration object for the Bitcoin chain, including network type and providers.
+ * @returns {Promise<number>} The estimated transaction fee in satoshis.
+ */
+export const getEstimatedFeeBTC = async (transaction: {
```
hcho112, line 153: Just like `signAndSend` and `getDerivedAddress`, can we also wrap `getEstimatedFeeEVM` and `getEstimatedFeeBTC` into a single function with switch case logic?

If you are concerned about maintaining the gas fee unit, we can also return like:
{ type: 'gwei', totalTransactionFee: NUMBER }

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
```diff
@@ -0,0 +1,24 @@
+export interface FunctionCall {
```
hcho112, line 1: we have this type defined on `import { FunctionCall } from '@near-js/transactions'` lets reference from original if possible?

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    method_name: string;
+    args: string;
+    gas: number;
+    deposit: string;
+  };
+}
+
+export type Actions = FunctionCall;
+
+export interface DelegateActionRelayerFormat {
```
hcho112, line 12: Same for `import { DelegateAction } from '@near-js/transactions';`

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -0,0 +1,152 @@
...
+  contract: ChainSignatureContracts
+): Promise<RSVSignature> => {
+  const functionCall = transactions.functionCall(
+    'sign',
+    {
+      payload: Array.from(ethers.getBytes(transactionHash)).slice().reverse(),
+      path,
+    },
+    new BN('300000000000000'),
+    new BN(0)
```
hcho112, line 46: These are repeated, lets pull them out into constants above with appropriate name to give a bit more context what they are?
`NEAR_MAX_GAS` or something similar

hcho112 (commented):
Left few minor improvements

packages/near-fast-auth-signer/src/components/Test/test.tsx
Pessina: Yes, for sure, forgot to clean up. Good catch!

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
Pessina: Good point... I'm not sure if mainnet has the same problem. I only tested on testnet as mainnet need real funds. 

I will update the comment

packages/near-fast-auth-signer/src/utils/multi-chain/chains/Bitcoin.ts
Pessina: Why do you need the transaction in the `fast-auth-signer` and encode it with base64?

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
Pessina: The types are different

This is the definition of the FunctionCall on '@near-js/transactions'

`export declare class FunctionCall extends IAction {
    methodName: string;
    args: Uint8Array;
    gas: BN;
    deposit: BN;
}`

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
Pessina: Same reason as the previous comment, the types are different

`export declare class DelegateAction extends Assignable {
    senderId: string;
    receiverId: string;
    actions: Array<Action>;
    nonce: BN;
    maxBlockHeight: BN;
    publicKey: PublicKey;
}`

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
Pessina: Agreed, will update it :)

packages/near-fast-auth-signer/src/utils/multi-chain/multiChain.ts
Pessina: That's a good point that me and @esaminu discussed, we can have a discussion on this.

But the main issues with a single method are:

- Complexity: As more chains are added, the function's argument types and body will become increasingly complex.
- Usability: Without TypeScript, it will be challenging for callers to use and understand the function correctly, making documentation and usage harder.

hcho112: > Complexity: As more chains are added, the function's argument types and body will become increasingly complex.
Wouldn't EVM ones will have similar types? even if we have more and more support on different coins, at some point many of them will overlapp. 

> Usability: Without TypeScript, it will be challenging for callers to use and understand the function correctly, making documentation and usage harder.
We could potentially do like `Arg type = type A | type B ....` and apply types separately on calling dedicated functions inside switch statement.

As we discussed, at the end of the day, would be nicer if user just simply calls top level function and get the fee data that they want. But if this is too time consuming, I'm happy to go with what we have

Pessina: Fully agree with calling only top level function. But what we are afraid it's that this functions will get very complex. 

I'm open to set a call and talk about it, maybe a single function it's idd better. 

I will create a ticket for it and we discuss for the module creation

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
hcho112: hmm, I thought we are constructing FunctionCall to do the same job.

hcho112: Just curiosity, when you say its different, I did notice that its `method_name` is different from `methodName` and `args` are in string. At the end, are we doing the same thing as classic signDelegate from NAJ? If we are doing the same thing shouldn't we fix the case issue? Or have completely different name for these Types? (So that users won't get confused why `FunctionCall` from NAJ VS `FunctionCall` from fast-auth-signer?

packages/near-fast-auth-signer/src/utils/multi-chain/relayer/types.ts
hcho112: same as above

hcho112 (approved):
LGTM, great work!!

PR #193, opened 2024-03-16
Refactor SignMultichain component and improve type checking

Pessina:
- Refactor SignMultichain component to handle message events more efficiently
- Improve type checking for Chain and related types
- Simplify signMultichainTransaction function and remove unnecessary useEffects
- Update utility functions to use the improved Chain type
- Clean up and optimize code in various files

### DIFF
--- a/packages/near-fast-auth-signer/src/components/Login/Login.tsx
+++ b/packages/near-fast-auth-signer/src/components/Login/Login.tsx
@@ -54,7 +54,7 @@ function Login() {
     data: { email: string }
   ) => {
     const newParams = new URLSearchParams(currentSearchParams);
-    newParams.set('email', encodeURIComponent(data.email));
+    newParams.set('email', data.email);
     navigate({
       pathname: '/add-device',
       search:   newParams.toString(),
--- a/packages/near-fast-auth-signer/src/components/Login/Login.tsx
+++ b/packages/near-fast-auth-signer/src/components/Login/Login.tsx
@@ -54,7 +54,7 @@ function Login() {
     data: { email: string }
   ) => {
     const newParams = new URLSearchParams(currentSearchParams);
-    newParams.set('email', data.email);
+    newParams.set('email', encodeURIComponent(data.email));
     navigate({
       pathname: '/add-device',
       search:   newParams.toString(),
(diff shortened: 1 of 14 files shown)


packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -267,7 +245,7 @@ function SignMultichain() {
           size="large"
           label={inFlight ? 'Loading...' : 'Approve'}
           onClick={onConfirm}
-          disabled={inFlight || !isValid || typeof amountInfo.price !== 'number'}
+          disabled={inFlight || !isValid || !firebaseUserLoading || !firebaseUser || typeof amountInfo.price !== 'number'}
```
hcho112, line 248: `!firebaseUserLoading` -> `firebaseUserLoading` need to be changed

Pessina: True, updated :)

packages/near-fast-auth-signer/src/components/Login/Login.tsx
```diff
@@ -54,7 +54,7 @@ function Login() {
     data: { email: string }
   ) => {
     const newParams = new URLSearchParams(currentSearchParams);
-    newParams.set('email', encodeURIComponent(data.email));
```
hcho112, line 57: I made this fix because I noticed that if we insert email like:
test+1@gmail.com, previous implementation actually returns `test @gmail.com`. with `encodeURIComponent` it actually returns `test+1@gmail.com`. When you revert this change, is there a reason why you have reverted this?

Pessina: Hmmm... then we have a problem. 

It was double encoding so we would need to double decode. 

So the user add it's email to Login, but when he reached AddDevice, the email was partially decoded, so he didn't automatically proceed, he had to fix his e-mail. 

It was only happening when testing the fast-auth-signer alone, if it was tested with near-discovery it didn't happen

Weird bug, I can revert it and we fix it later, I thought was a quick fix. But it not related to the PR

hcho112 (approved):
LGTM

PR #197, opened 2024-03-19
feat: make `relayerUrl` optional

esaminu:
Make `relayerUrl` optional on the multichain toolkit module so that it can be used with non fast auth accounts. This is necessary because the meta transaction relayer will only cover gas for fast auth accounts:

> For the fastauth relayer, the account needs to have an allowance in redis initialized by calling /create_account_atomic . This means that unless the account has been registered, the fastauth meta transaction relayer will not cover gas for any account

### DIFF
--- a/packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
+++ b/packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
@@ -29,8 +29,8 @@ const toRVS = (signature: string): RSVSignature => {
  * @param {string | ethers.BytesLike} transactionHash - The hash of the transaction to be signed.
  * @param {string} path - The derivation path used for signing the transaction.
  * @param {NearAuthentication} nearAuthentication - The NEAR accountId, keypair and networkId used for signing the transaction.
- * @param {string} relayerUrl - The URL of the relayer service to which the signed transaction is sent.
  * @param {ChainSignatureContracts} contract - The contract identifier for chain signature operations.
+ * @param {string} relayerUrl - The URL of the relayer service to which the signed transaction is sent.
  * @returns {Promise<RSVSignature>} A promise that resolves to the RSV signature of the signed transaction.
  * @throws {Error} Throws an error if the signature cannot be fetched after 3 attempts.
  */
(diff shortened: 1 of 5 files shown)


packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -29,28 +29,18 @@ const toRVS = (signature: string): RSVSignature => {
...
  * @throws {Error} Throws an error if the signature cannot be fetched after 3 attempts.
  */
 export const sign = async (
   transactionHash: string | ethers.BytesLike,
   path: string,
   nearAuthentication: NearAuthentication,
-  relayerUrl: string,
-  contract: ChainSignatureContracts
+  contract: ChainSignatureContracts,
+  relayerUrl?: string
```
hcho112, line 42: since sign function now accepts 5 arguments, would be ideal to change the argument with object with props. In that way order doesn't really matter and reduces chance of making mistake.

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -74,6 +64,46 @@ export const sign = async (
...
+    return {
+      r: R.slice(2),
+      s
+    };
+  }
+
+  const functionCall = actionCreators.functionCall(
+    'sign',
+    {
+      payload: Array.from(ethers.getBytes(transactionHash)).slice().reverse(),
```
hcho112, line 100: `Array.from(ethers.getBytes(transactionHash)).slice().reverse()` used twice, lets pull it out to a constant?

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -74,6 +64,46 @@ export const sign = async (
 
   const account = new Account(connection, nearAuthentication.accountId);
 
+  if (!relayerUrl) {
+    const multichainContractAcc = new Contract(account, contract, {
```
Pessina, line 68: The contract it's defined twice, one here and other for the pulic_key

packages/near-fast-auth-signer/src/utils/multi-chain/signature/signature.ts
```diff
@@ -74,6 +64,46 @@ export const sign = async (
...
+
+    const [R, s] = await multichainContractAcc.sign({
+      args: {
+        payload: Array.from(ethers.getBytes(transactionHash)).slice().reverse(),
+        path
+      },
+      gas: NEAR_MAX_GAS
+    });
+
+    return {
```
Pessina, line 91: re-use toRSV function

Pessina (changes requested):

hcho112 (approved):
LGTM

Pessina (approved):

PR #199, opened 2024-03-24
dApp Integration for SignMultichain Component

Pessina:
- Update SignMultichain component to handle incoming messages from dApps
- Improve error handling and message passing between iframe and parent window
- Refactor code for better readability and maintainability
- Fix issues with key derivation and transaction signing
- Add support for passing key version in contract calls

### DIFF
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -25,7 +25,7 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
     }
 
     try {
-      const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
+      const res = await withTimeout(fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION);
       if (res) {
         const ids = await res.json();
         return ids;
@@ -54,15 +54,8 @@ export const fetchAccountIdsFromTwoKeys = async (
   keyPairA: KeyPair,
   keyPairB: KeyPair
 ): Promise<{ accId: string, keyPair: KeyPair}> => {
-  const accountIdsFromPublicKeyA = await withTimeout(
-    fetchAccountIds(keyPairA.getPublicKey().toString()),
-    KIT_WALLET_WAIT_DURATION
-  );
-
-  const accountIdsFromPublicKeyB = await withTimeout(
-    fetchAccountIds(keyPairB.getPublicKey().toString()),
-    KIT_WALLET_WAIT_DURATION
-  );
+  const accountIdsFromPublicKeyA = fetchAccountIds(keyPairA.getPublicKey().toString());
+  const accountIdsFromPublicKeyB = fetchAccountIds(keyPairB.getPublicKey().toString());
 
   const firstResult = await Promise.race([accountIdsFromPublicKeyA, accountIdsFromPublicKeyB]);
   const firstKey = firstResult === await accountIdsFromPublicKeyA ? keyPairA : keyPairB;
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -25,7 +25,7 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
     }
 
     try {
-      const res = await withTimeout(fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION);
+      const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
       if (res) {
         const ids = await res.json();
         return ids;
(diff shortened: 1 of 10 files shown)


packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -25,7 +25,7 @@ export const fetchAccountIds = async (publicKey: string): Promise<string[]> => {
     }
 
     try {
-      const res = await withTimeout(fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`), KIT_WALLET_WAIT_DURATION);
+      const res = await fetch(`${network.fastAuth.authHelperUrl}/publicKey/${publicKey}/accounts`);
```
hcho112, line 28: would like to know why withTimeout is removed on this line. Is it more preferred to wait longer until we get a response back from kit wallet?

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -108,7 +98,7 @@ function SignMultichain() {
         feeProperties
       });
       if (response.success) {
-        window.parent.postMessage({ type: 'response', message: `Successfully sign and send transaction, ${response.transactionHash}` }, '*');
+        window.parent.postMessage({ type: 'multiCHain', message: `Successfully sign and send transaction, ${response.transactionHash}` }, '*');
```
hcho112, line 101: `multiCHain` -> `multiChain`

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -73,7 +63,7 @@ function SignMultichain() {
   });
 
   const onError = (text: string) => {
-    window.parent.postMessage({ type: 'multichainError', message: text }, '*');
+    window.parent.postMessage({ type: 'multiChain', message: text }, '*');
```
hcho112, line 66: I don't mind the change, but would it be ideal to indicate that if the passing message is error or success response? so that receiving end can apply appropriate decision around given message? (eg, if error, show red/orange colour with pop message?)

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -172,14 +162,11 @@ function SignMultichain() {
       handleMessage,
     );
 
-    // TODO: test code, delete later
-    // handleMessage({ data: { type: 'multi-chain', data: sampleMessageForBTC } });
+    window.parent.postMessage({ type: 'sign-multichain-load', message: 'SignMultichain page has loaded' }, '*');
```
hcho112, line 165: different type convention, we should keep the consistency on use kebab-case or camelCase

packages/near-fast-auth-signer/src/hooks/useAuthState.ts
```diff
@@ -42,7 +42,7 @@ export const getAuthState = async (email?: string | null): Promise<AuthState> =>
         networkId,
       });
 
-      await window.fastAuthController.setKey(new KeyPairEd25519(accountInfo.publicKey.split(':')[1]));
+      await window.fastAuthController.setKey(new KeyPairEd25519(accountInfo.keyPair.toString().split(':')[1]));
```
hcho112, line 45: too many logic in one line, can we fix it so that if accountInfo is not set, we don't run this part of code?
eg
```
if (accountInfo && accountInfo.keyPair) {
  const keyPairParts = accountInfo.keyPair.toString().split(':');
  if (keyPairParts.length > 1) {
    await window.fastAuthController.setKey(new KeyPairEd25519(keyPairParts[1]));
  }
}
```

hcho112 (changes requested):
left few minor small fixes

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
Pessina: Good point, changing it :)

packages/near-fast-auth-signer/src/hooks/useAuthState.ts
Pessina: We already check for accountInfo here: https://github.com/near/fast-auth-signer/blob/0d40cd0318793b59ebd93f2373c41fc618a611f9/packages/near-fast-auth-signer/src/hooks/useAuthState.ts#L36

I would prefer to do it in one line, it's just string operation.

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
Pessina: I think we can discuss this, we can provide it on the type or the message body (e.g. making it an object with more descriptive information. 

It's the wallet that will handle this incoming message. 

It's ok to do it in another PR?

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
Pessina: Good point, changing :)

packages/near-fast-auth-signer/src/api/index.ts
Pessina: I think I forgot to rollback this code after my debugging. I will revert this changes. Good catch! :)

packages/near-fast-auth-signer/src/hooks/useAuthState.ts
hcho112: but above is for checking `accId` but I'm more concerned about `.keyPair` is not set.

Pessina: I'm not seeing why it could break, typescript it's not showing me any error here. 

I will add the ?. for safety. 

But, if the accountInfo exist, the KeyPair exist, no? The KeyPair come from the passkeys, and we assign the correct KeyPair based on which one is the real one.

hcho112 (dismissed):
LGTM

hcho112 (approved):

PR #205, opened 2024-03-28
Multichain UI flow for Domain Key and Wrong Domain Key

hcho112:
This PR contains implementation for UI of Domain Key and Wrong Domain Key.
[Design Ref](https://www.figma.com/file/jB85CJOUewdvpB1e7wCLwO/Multi-chain?node-id=2957%3A14344&mode=dev)

For Domain key flow:
Safari will have its own UI due to its security policy around passkey modal requires user action:

https://github.com/near/fast-auth-signer/assets/6027014/8fe69f85-de49-433a-b09b-1c622881fefa

Non Safari browsers will have UI hidden and all the process will be automatic after passkey confirmation

https://github.com/near/fast-auth-signer/assets/6027014/004240e1-b6fe-4597-8430-b8e4e502dc5a

For Wrong domain key flow:

image
They can only proceed after click the check box

This PR requires https://github.com/near/near-fastauth-wallet/pull/22 to be also merged and available


packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -138,6 +181,9 @@ function SignMultichain() {
           setValid(true);
           setIsDomainKey(transaction?.domain === event?.origin);
           setMessage(transaction);
+          if (transaction?.domain === event?.origin && !isSafariBrowser) {
```
Pessina, line 184: Duplicated `if (transaction?.domain === event?.origin && !isSafariBrowser) {`

I think one approach we can take it's hide the modal until we need to display it. 

So, instead of notify when we need to hide the modal, we notify the caller when we need to display it. 

What do you think? So we can do all the operations inside a single `if`

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -172,34 +218,37 @@ function SignMultichain() {
...
-        </div>
+        </Container>
         {error && <p className="info-text error">{error}</p>}
       </ModalSignWrapper>
     );
   }
 
   return (
-    <ModalSignWrapper ref={signTransactionRef}>
+    <ModalSignWrapper ref={signTransactionRef} hide={message?.domain === origin ? 'true' : undefined} warning={isUnsafe}>
```
Pessina, line 245: Why not a boolean for the hide? 

Then you could do only message.domain === origin

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
```diff
@@ -19,7 +20,11 @@ import { getAuthState } from '../../hooks/useAuthState';
 import useIframeDialogConfig from '../../hooks/useIframeDialogConfig';
 import InternetSvg from '../../Images/Internet';
 import ModalIconSvg from '../../Images/ModalIcon';
+import WarningCircle from '../../Images/WarningCircle';
```
Pessina, line 23: I think we have a inconsistent naming here. As you can see the `InternetSvg` and `ModalIconSvg`are named with SVG at the end.

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: yeah, but on console, it was showing warning about mounting boolean on HTML element, with saying that it suggests to apply above solution

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: I tried and I ended up duplicating the if statement because if I use it under single if statement, it does show the modal for some period of time (fairly broken) until the rest begins. 

To give you a bit more context, there are two logic to consider:
1. UI hide logic
2. Executing `signMultichainTransaction` 

The reason why I didn't put them in single if statement is because there is async operation of fetching prices before calling `signMultichainTransaction`. So if I wrap them both under one if statement and if it is located after price fetch request, there is a broken UI happens. I can't put if statement with both logic before price fetch because it won't work at all. 

So it may look like if statement is duplicated, but the way I see it is that it is optimised for UX.

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
hcho112: nvm, I have reverted into boolean

Pessina (approved):

packages/near-fast-auth-signer/src/components/SignMultichain/SignMultichain.tsx
Pessina: Thanks Danny :)

PR #207, opened 2024-03-29
fix: avoid multiple passkey prompts and sign in event logs

esaminu:
Adds:

- On click of the email address input on the login form, prompt for passkey. If provided log the user in immediately if not send email after they fill in the address and click the submit button

Fixes:

-  Remove duplication of event `click-login-continue`. Duplication happens when email is send due to manual `onSubmit` trigger after redirect
- Passkey is prompted twice when email is sent currently due to no longer using `webauthn_username`

### DIFF
--- a/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
+++ b/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
@@ -371,7 +371,11 @@ function AddDevicePage() {
             input: 'add-device-email',
           }}
           onFocus={async () => {
-            if (!wasPassKeyPrompted && await isPassKeyAvailable()) {
+            if (
+              !wasPassKeyPrompted
+              && decodeIfTruthy(searchParams.get('forceNoPasskey')) !== true
+              && (await isPassKeyAvailable())
+            ) {
               setInFlight(true);
               const authenticated = await getAuthState();
               setWasPassKeyPrompted(true);
--- a/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
+++ b/packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
@@ -26,9 +26,34 @@ import {
 import { recordEvent } from '../../utils/analytics';
 import { basePath } from '../../utils/config';
 import { checkFirestoreReady, firebaseAuth } from '../../utils/firebase';
+import ErrorSvg from '../CreateAccount/icons/ErrorSvg';
 import { FormContainer, StyledContainer } from '../Layout';
 import { Separator, SeparatorWrapper } from '../Login/Login.style';
 import { getMultiChainContract } from '../SignMultichain/utils';
+
+const ErrorContainer = styled.div`
+.stats-message {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+
+  span {
+    flex: 1;
+  }
+
+  svg {
+    flex-shrink: 0;
+  }
+
+  &.error {
+    color: #a81500;
+  }
+
+  &.success {
+    color: #197650;
+  }
+}
+`;
 
 export const handleCreateAccount = async ({
   accountId, email, isRecovery, success_url, failure_url, public_key, contract_id, methodNames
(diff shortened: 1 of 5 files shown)


packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -269,19 +269,22 @@ function AddDevicePage() {
   }, [firebaseUser, navigate, searchParams]);
```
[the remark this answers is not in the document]
Pessina: Also I'm getting this error

https://github.com/near/fast-auth-signer/assets/22358004/d57b823e-1bca-4cef-ba24-acab5dd3f563

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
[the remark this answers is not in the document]
esaminu: I don't see the connection between the linked PR and the points this PR addresses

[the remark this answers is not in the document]
esaminu: Fixed the issue, it was attempting to access Firestore when there is no auth state

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -84,6 +84,7 @@ function AddDevicePage() {
   const defaultValues = {
```
Pessina, line 84: Bug: User login with e-mail on input instead of email used on passkeys


https://github.com/near/fast-auth-signer/assets/22358004/dfdacd95-6f4d-44c5-98d0-ba18d09734f3

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
```diff
@@ -84,6 +84,7 @@ function AddDevicePage() {
...
     email: searchParams.get('email') ?? '',
```
Pessina, line 85: Not major, but I think we need to display a message. 

If the user remove all keys from the account on Remove Keys Screen, then try to login again with the e-mail he deleted the keys, the Login will fail but there is no error message. 

Then, if the user, do the flow with the email it will work fine (He goes to verify e-mail and it works)

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
esaminu: This will only happen if the email is pre filled. It's expected in this case because the user did not click Continue they clicked the input to edit the email

Pessina, line 84: Loop on remove keys

https://github.com/near/fast-auth-signer/assets/22358004/376cfb3c-68ce-43f2-814a-7c2a41c2ac1a

Pessina (changes requested):

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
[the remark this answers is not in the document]
Pessina: Cool, thanks :)

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
Pessina: But is this ok? The user select one passkey but endup on other account... I think it's a security problem, no? 

Imagine if 3 people are sharing the same computer and one can login into the other

esaminu: This is not possible unless the keypair for the selected account belongs to the first account

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
esaminu: The fix for this is to not allow the user to delete the current signer instance's FAK. Can be addressed in a follow up

packages/near-fast-auth-signer/src/components/AddDevice/AddDevice.tsx
esaminu: Removed pre filling of email to address this

Pessina (approved):

hcho112 (approved):
LGTM

benmalcom (approved):

PR #210, opened 2024-04-02
feat: add support for docker container envs by supporting runtime env variables

esaminu:
Support environment variable injection at compile-time and runtime. This is required so that our docker image can be run with custom environment variables which currently does not work since our image is already bundled with env variables that were passed at build time.

Testing Instructions:
- Build and run the Docker container.
- Pass necessary environment variables at runtime (e.g. using `-e` flag).
- Confirm that you are able to override defaults provided by webpack and at build time


hcho112 (approved):
I have tested with following commends:
```
docker build --no-cache -t tagname ../.. --file Dockerfile --progress plain
docker run -e ENV1=VAL1  -p 8080:80 tagname
```
Where ENV1 can be one of environment variables on template file
Then go to port 8080 on localhost and make sure ENV1 is successfully overwritten

I have confirmed that window.env.XXXX has been overwritten when I modify on docker commend

hcho112:
I would recommend to update readme or create a new readme for docker only and briefly note above to apply environment variable?

hcho112 (approved):
Good read me update, LGTM

PR #214, opened 2024-04-04
Initial implementation of fast-near on fetchAccountIds

hcho112:
This PR contains implementation for introducing fastnear on mainnet only.

[As per issue](https://github.com/fastnear/fastnear-api-server-rs/issues/2), unfortunately it is currently supporting mainnet environment only. 

Currently some of accounts on near.org fail to login due to issue on kitwallet. This PR will resolve authentication issues on mainnet/near.org.

I have tested following on Chrome browser:
Mainnet:
1. Authenticate with existing account (success)
2. Create new account (success)
3. Authenticate with account created above (success)

Testnet:
Same flow, however confirmed that logic implemented on PR is not executed as per design.


hcho112:
Closing this PR due to https://github.com/near/fast-auth-signer/pull/216

PR #216, opened 2024-04-08
Initial implementation of queryAPI on fetchAccountIds

hcho112:
This PR contains implementation for introducing queryAPI on mainnet only.

[Demo on retrieving account id](https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=dataplatform.near/access_keys_v1) 
Click on “Show GraphiQL Code Exporter” icon to get the JS code snippet.

Currently some of accounts on near.org fail to login due to issue on kitwallet. This PR will resolve authentication issues on mainnet/near.org.

I have tested following on Chrome browser:
Mainnet:
1. Authenticate with existing account (success)
2. Create new account (success)
3. Authenticate with account created above (success)

Testnet:
Same flow, however confirmed that logic implemented on PR is not executed as per design.

### DIFF
--- a/packages/near-fast-auth-signer/src/api/index.ts
+++ b/packages/near-fast-auth-signer/src/api/index.ts
@@ -29,7 +29,7 @@ const fetchAccountIdFromQueryApi = async (publicKey: string): Promise<string | n
         operationName: 'AccountIdByPublicKey',
       }),
     });
-    if (res) {
+    if (res.ok) {
       const response = await res.json();
       if (response.data) {
         const data = response.data.dataplatform_near_access_keys_v1_access_keys_v1[0];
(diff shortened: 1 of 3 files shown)


packages/near-fast-auth-signer/src/api/index.ts
```diff
@@ -2,11 +2,47 @@ import { captureException } from '@sentry/react';
...
+    const res = await fetch(network.fastAuth.queryApiUrl, {
+      method:  'POST',
+      headers: { 'x-hasura-role': 'dataplatform_near' },
+      body:    JSON.stringify({
+        query,
+        variables:     {},
+        operationName: 'AccountIdByPublicKey',
+      }),
+    });
+    if (res) {
```
benmalcom, line 32: @hcho112 I suggest checking for `res.ok` here instead of `res`, as `res` will always be truthy even for error responses. Using `res.ok` ensures that we specifically verify the success of the HTTP request, based on its status code. This helps to handle both successful and error responses appropriately.

benmalcom (approved):

PR #223, opened 2024-04-30
Implement Fallback to User Balance for Transactions When Relayer Fails

Pessina:
This PR introduces a fallback mechanism for transactions when the relayer service fails to relay them due to specific issues such as actions not being allowed or the sender/receiver not being whitelisted. In such cases, the transaction will be retried using the user's own balance, ensuring that the transaction can still be processed without dependency on the relayer's current state or configuration.


hcho112:
While I was looking at the code, it comes to my mind that, would it be nicer if:
1. We inform the user (or developer of this code or both) that upon relayer fail, we will attempt to send transaction with user balance.
2. Toggle switch logic (either via UI or config via developer) where user agrees to execute the transaction via their account balance. 

TDLR, I don't think just assume that user wants to transaction with their account balance is a good thing. It is possible that either they may not want this feature nor not being aware of this will happen. (Some user may even think this is a scam because they may expect using fast-auth + relayer would pay the gas but it ends up using user's balance)

Ideally, we should implement both where:
1. upon configuration/implementation, if we pass `allowToUseUserBalance` or some sort of boolean flag. 
2. If above boolean flag is set to true and if it is multi-chain, we show small UI boolean flag with description about "Allow to use user balance to send transaction".
3. If signTransaction is called and relayer is failed, we inform user in some way? (show UI/modal/popup/notification etc)

and need to rebase and fix yarn lock file

Pessina:
@hcho112 Good point!

I partially agree with you.

For the developer part, I completely agree and will include the configuration variable that you mentioned.

However, for the user part, I believe they are already aware of the fees as it's already stated on the iFrame of the fast-auth-signer.

Screenshot 2024-05-02 at 2 40 53 PM

As you can see, the total already includes the fees. Users assume they will pay the gas, but sometimes they won't. It's a positive surprise in that case.

Pessina:
@hcho112 regarding the multi-chain, this PR doesn't touched multi-chain code, it's only for the `/sign` method.

hcho112:
@Pessina 
Personally I think this is a good change, I would rather have something that works even if I need to pay transaction with my account. However I would rather confirm with @diogoortega and @nall-near for further feedback. Can we get your feedback?

In terms of making change on `near-fastauth-wallet`, I don't think it needs to be mentioned in NEP. We can make the config changes by updating `FastAuthWalletParams` type with adding extra prop and pass down from there?

diogoortega:
Agree that it's great to have a fallback mechanism instead of just having the transaction fail.
However, we should inform the user. When will we know if the transaction will be "sponsored" (via relayer) or not? Can we show that info in the confirmation dialog [linked above](https://private-user-images.githubusercontent.com/22358004/327351808-10c233c4-5280-4197-a4f9-d6a969eac4d7.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTQ3NTE5NjksIm5iZiI6MTcxNDc1MTY2OSwicGF0aCI6Ii8yMjM1ODAwNC8zMjczNTE4MDgtMTBjMjMzYzQtNTI4MC00MTk3LWE0ZjktZDZhOTY5ZWFjNGQ3LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MDMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTAzVDE1NTQyOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTE1YjBmNTdmOTJhMjY1YjJhZTMzYjRjMzhkYjJhMzRlOWVjMDJkZTg3OTQ1MzY1ZjdjNGRjNWE2ZjU5MDgyMjEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.1UhjASmK7CMNMU4BMwhOj28Jcfcip1qs5v5m6qxvKAQ) or would we only know after the transaction is confirmed by the user?

If we don't know beforehand, maybe we could add a checkbox (default ON) that says something like "Use my balance for this transaction if sponsoring is not available". Similar to this toggle on iOS that tries to send your text via SMS when iMessage isn't available: 

IMG_9836

What do you think @nall-near?

nall-near:
I think the context of _why_ the transaction fails is important. For example, in the case of un-allowed actions, or non-whitelisted receiver ID, these hurdles could be intentional to protect the user from carelessly signing transactions. 

Similar to Dio's question above,
> When will we know if the transaction will be "sponsored" (via relayer) or not?
Will we know that the proposed transaction does not meet the appropriate criteria (whitelist, allowed, etc.) before signing?

Regardless, I agree with @diogoortega that we need to inform the user if something is different/unexpected about signing, and try to be as explicit as possible.
A. Txn fails because no more gas allotted | Inform user that they will need to pay for the fees
B. Txn fails because the receiver is not whitelisted | Inform the user this is not inherently trusted and they should proceed with caution
C. Txn fails because the action is not allowed | similar to B., inform the user that this action is not automatic and to confirm before continuing

Let me know if I'm overthinking this, we don't need unique messaging for each use case, but I do want to alert the user with some caution if they are trying to do something outside the allowed context of the relayer as well as when they may need to pay for the fees themselves.

I also like Dio's suggestion to have a toggle within the dapp context settings that allows users to confirm automatically.

Pessina:
@diogoortega and @nall-near thanks for all the input, let me try answer your questions, let me know if it's not clear. 

> When will we know if the transaction will be "sponsored" (via relayer) or not?

Currently all the transactions sent by the `fast-auth-wallet` are sent via relayer and we don't know if they will fail before calling the relayer. So, currently, the only way to know if the transaction will be "sponsored" it's call the relayer and check the response of the call. 

I talked with @DavidM-D, please correct if I'm wrong, and we could implement the validation on FrontEnd. But I'm not sure how long it will take to be implemented, depends on the size of the logic, also we would have to keep both relayer/FE logic in sync with any update on relayer validation logic. I think we would need further discussion here, maybe an new endpoint on the relayer to validate the transaction... 

Those are the flows: 

**FullAccess Key:** 

1 - Current flow: 

dApp send tx to `fast-auth-wallet` -> `fast-auth-wallet` call `fast-auth-signer` asking signature -> `fast-auth-signer` iFrame pop up and user click on `Confirm` to sign the `SignedDelegate` -> `fast-auth-signer` return the `SignedDelegate` to `fast-auth-wallet` -> `fast-auth-wallet` send the transaction through the relayer -> transaction success of fail from relayer. 

2 - New flow

For the new flow, we need another type of signature, 




> Will we know that the proposed transaction does not meet the appropriate criteria (whitelist, allowed, etc.) before signing?

Pessina:
I just had a chat with @DavidM-D about this issue, and he came up with a nice suggestion that we think may simplify the 
flow.

The dApp probably knows if the transaction can or cannot be relayed, so we could expose two endpoints for the transaction signature:

- Relayer endpoint (signs DelegateAction)
- User balance endpoint (signs Transaction)

By doing this, we can create two distinct user interfaces to communicate the transaction details to the user, rather than having a single interface that attempts to handle both scenarios with a toggle box or by prompting the iFrame twice for confirmation.

What are your thoughts on this approach?

Also, I asked him how hard it would be to implement the feature to check if the relayer will accept/refuse the transaction before sending it. He said it will be complicated to do that, so it would be better if we avoid this path.

esaminu:
The solution to expose 2 endpoints is the most technically sound and makes the most sense. We should move the current implementation into a new `signAndSendDelegateAction` in `near-fastauth-wallet` and the signing of transactions and sending directly to rpc should be done in the current `signAndSendTransaction`. 

We will have to make changes to https://github.com/NearSocial/VM to call the `signAndSendDelegateAction` when a widget wants to use the relayer (It currently always calls `signAndSendTransaction`). 

As far as UX we can now just use different copy for the 2 different routes unless there are other suggestions

nall-near:
Hey @esaminu @Pessina, I agree with the two end points. we will keep the UX simple. always displaying the fee number but in the relayer instance mentioning that they are _paid_ or _sponsored_ and maybe crossing our the fee amount.

I'm still curious about the other cases regarding failed transactions and if we need to mitigate those eventually, but fine to move forward with your suggestion for now.

diogoortega:
Also agree with the two endpoints solution. Thank you for the additional details.

hcho112 (approved):
Again, awesome work @Pessina 
LGTM!

hcho112:
One thing to note, would be nicer if we start writing test for two different route?
@esaminu created PR for signTransaction so would be nicer if we create test once this PR is in master

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -205,6 +209,54 @@ class FastAuthController {
...
+    const { accessKey } = accessKeyInfo;
+
+    const block = await this.connection.provider.block({
+      finality: 'final',
+    });
+    const blockHash = block.header.hash;
+
+    const nonce = accessKey.nonce.add(new BN(1));
+
+    if (isPassKeyAvailable) {
```
esaminu, line 234: Do you mean to call this?

Pessina: Idd, updated :)

packages/near-fast-auth-signer-e2e-tests/test-app/App.tsx
```diff
@@ -74,7 +74,7 @@ export default function App() {
       )}
       <button
         type="button"
-        data-test-id="sign-transaction-button"
+        data-testid="sign-transaction-button"
```
hcho112, line 77: this is good fix, noticed that we have lots of places need to be fixed. (Should address in different PR)

hcho112 (commented):
Code looks good, wanted to test this out but we can't at the moment due to indexer issue. 

Meanwhile, wonder if it is possible to have a end to end test covering the new sign and sign-transaction route?

packages/near-fast-auth-signer/src/lib/controller.ts
```diff
@@ -205,6 +209,54 @@ class FastAuthController {
...
+        actions,
+        baseDecode(blockHash),
+        this.connection.signer,
+        this.accountId,
+        this.connection.networkId
+      );
+    }
+
+    const oidcToken = await firebaseAuth.currentUser.getIdToken();
+    const localKey = await this.getKey(`oidc_keypair_${oidcToken}`) || await this.getLocalStoreKey(`oidc_keypair_${oidcToken}`);
```
esaminu, line 247: This is the FRP key so it cannot be used to sign anything. If we want to implement fallback to signing with the oidc token then we need to make a request to the mpc service. IMO we can consider treating sign routes as iframe only and if we do then there is no need for fallback to oidc because iframes will not have access to that either way

packages/near-fast-auth-signer-e2e-tests/test-app/App.tsx
esaminu: Both are ok to use but I think we should standardize one across the codebase

packages/near-fast-auth-signer/src/lib/controller.ts
Pessina: Agree, I will remove the fallback code.

packages/near-fast-auth-signer-e2e-tests/test-app/App.tsx
Pessina: Yeah, we need a standard. I started cleaning up to use the Playwright way, that's `data-testid`.

packages/near-fast-auth-signer-e2e-tests/test-app/App.tsx
```diff
@@ -74,7 +74,7 @@ export default function App() {
...
         onClick={() => {
           fastAuthWallet
             .signAndSendTransaction(
```
esaminu, line 80: should we update this to the new `signAndSendSignedDelegate`? Tests should fail if we send directly to rpc because the created test account has no balance right?

packages/near-fast-auth-signer/src/components/Sign/Sign.styles.ts
```diff
@@ -7,127 +7,143 @@ interface ModalSignWrapperProps {
...
   display: flex;
   flex-direction: column;
   position: relative;
-  row-gap: 8px;
+  overflow: hidden;
+
+  .modal-body {
+    height: 100%;
+    width: 100%;
+    overflow: scroll;
```
hcho112, line 23: replace this to `auto` so that when it is collapsed, UI looks better as scroll is not needed. 
If you change this to auto:
image

Original code:
image

hcho112 (changes requested):
@Pessina I have tested them according to your recommendation but I couldn't get any of them working properly. Maybe perhaps we should try to test this together? meanwhile can you share with us video recording of working version if you can get this working locally?

image

image

packages/near-fast-auth-signer-e2e-tests/test-app/App.tsx
Pessina: This will be done once we publish the new wallet version. 

Currently the `signAndSendTransaction ` still send a `signedDelegate`and the test is passing as you can see on the CI

packages/near-fast-auth-signer/src/components/Sign/Sign.styles.ts
Pessina: Good point, I will replace it :)

Pessina:
@hcho112 I believe the error you are facing it's because you didn't linked the local `near-fastauth-wallet` package, that's why it's not finding the methods. 

In order to do that you need to run `pnpm link [path to your near-fastauth-wallet build]` on `near-discovery` root folder, then you can paste the test code I shared with you on Slack and conduct the testing. 

Those are the recordings: 

https://github.com/near/fast-auth-signer/assets/22358004/6d2f0316-6187-4e17-9d73-cfc93925ab3b

https://github.com/near/fast-auth-signer/assets/22358004/0737a34d-5f73-45ae-836d-d447113f7d9a

If you still face issues, please let me know and we can work on this together.

hcho112 (dismissed):
LGTM!! awesome work

PR #225, opened 2024-04-30
Handle AuthCallback errors

benmalcom:
This PR handles a common error in `fast-auth` testing which is the firebase `auth/invalid-action-code` error, and also adds a generic error handling to the page. We should be able to show errors in the `AuthCallback` page instead of just redirecting users to the gateway url where they don't have much info on what happened.

To see this simple view  you can trigger any error, e.g clicking an already used firebase link from an email.

Example below

iPhone-13-PRO-localhost


hcho112:
@benmalcom code looks good, can we have some following info:
1. How would you manually create this error?
2. Screenshot of Error handling view

benmalcom:
> @benmalcom code looks good, can we have some following info:
> 
> 1. How would you manually create this error?
> 2. Screenshot of Error handling view

@hcho112 done

hcho112 (approved):
LGTM

PR #229, opened 2024-05-11
Auth test flow

Pessina:
Add e2e testing using playwright for the following flows: 

- Create account and login with email
- Create account and login with passkeys


packages/near-fast-auth-signer-e2e-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import PageManager from '../pages/PageManager';
+import { test } from '../test-options';
+import { getRandomEmailAndAccountId } from '../utils/email';
+import { rerouteToCustomURL } from '../utils/url';
+
+test.beforeEach(async ({ page, baseURL, relayerURL }) => {
+  await page.goto(baseURL);
+
+  // Overwrite relayer URLs to use local instance
```
esaminu, line 11: Why not use the `relayerUrl` param in `setupFastAuthWallet` or use [the `setRelayerUrl` method](https://github.com/near/near-fastauth-wallet/blob/main/src/lib/near-fastauth-wallet.ts#L346) here?

Pessina: My idea was to take advantage of the playwright projects: https://playwright.dev/docs/test-projects#configure-projects-for-multiple-environments

So we configure local, testnet, mainnet, etc... projects directly from `playwright.config.ts` without changing the App code. 

I think it's easier and more clear. What do you think?

hcho112 (commented):
LGTM! everything looks great, amazing work @Pessina I have very few comments about this PR, yet this wouldn't be a stopper. 

1. Lets update readme file with what environment variable need to be set in order to run the test successfully
2. Vercel still fails on this PR, do you know whats wrong with it?

hcho112:
Actually I will address readme on my PR, my PR is almost complete anyway.

Pessina:
Perfect @hcho112 , thank you for the review and updating the Readme. I will check your PR after we merge this one :)

benmalcom (approved):
Looks good @Pessina.

You can put some information in the README about the .env variables we use for the tests.

PR #244, opened 2024-06-05
E2e test for sign-multichain

benmalcom:
This is the PR for sign-multichain e2e tests


packages/near-fast-auth-signer-e2e-tests/mpc-service.ts
```diff
@@ -0,0 +1,153 @@
+import { KeyPair } from '@near-js/crypto';
```
Pessina, line 1: I see some parts of it are copied from the app source code. Can we import it instead of copy? So we don't need to maintain duplicated code.

packages/near-fast-auth-signer-e2e-tests/package.json
```diff
@@ -14,20 +14,24 @@
   ],
   "scripts": {
     "test": "yarn playwright test",
-    "start": "webpack serve --mode development --history-api-fallback"
+    "start": "webpack serve  --mode development --history-api-fallback"
   },
   "dependencies": {
+    "@covalenthq/client-sdk": "^1.0.2",
```
Pessina, line 20: I'm not sure if it's possible. 

But can we avoid adding test dependencies on the app dependencies? In order to avoid a big package.

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    await signMultiChain.clickApproveButton();
+    const multiChainResponse = await signMultiChain.waitForMultiChainResponse();
+    expect(multiChainResponse.transactionHash).toBeDefined();
+    expect(multiChainResponse).toHaveProperty('message');
+    expect(multiChainResponse.message).toBe('Successfully signed and sent transaction');
+    await expect(page.locator('#nfw-connect-iframe')).not.toBeVisible();
+    await expect(page.locator('#nfw-connect-iframe')).not.toBeVisible();
+  });
+
+  // Skipping this because of the unpredictable nature concerning replacement fee
```
Pessina, line 102: Shall we create a ticket for this?

packages/near-fast-auth-signer-e2e-tests/firebase.ts
```diff
@@ -0,0 +1,74 @@
+import admin from 'firebase-admin';
+
+import { serviceAccount } from './utils/serviceAccount';
+
+admin.initializeApp({
```
Pessina, line 5: I'm not sure, but I believe some of the functions for firebase are already create by @hcho112 . 

Like this, I think both do the same thing, no? https://github.com/near/fast-auth-signer/blob/b6bfb4516130c7de6798a5c3ec5b99d28b5d8fae/packages/near-fast-auth-signer-e2e-tests/utils/firebase.ts#L29

If that's the case, please re-use the ones from Danny and include yours on the same file.

Pessina (changes requested):

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
benmalcom: Yes please.

packages/near-fast-auth-signer-e2e-tests/firebase.ts
benmalcom: Both were actually created by @hcho112, this PR has nothing to do with both files, I only rebased after @hcho112's merge to be update.

Both files are already in `main`.

packages/near-fast-auth-signer-e2e-tests/mpc-service.ts
benmalcom: Removed. 
It's actually useless now, it's already imported.

packages/near-fast-auth-signer-e2e-tests/firebase.ts
hcho112: @benmalcom if you look at latest main, you will notice that firebase.ts is moved under utils/firebase.ts

so this file is redundant. If you have added new functions, these should be allocated to utils/firebase.ts

packages/near-fast-auth-signer-e2e-tests/package.json
```diff
@@ -14,20 +14,23 @@
   ],
   "scripts": {
     "test": "yarn playwright test",
-    "start": "webpack serve --mode development --history-api-fallback"
+    "start": "webpack serve  --mode development --history-api-fallback"
```
hcho112, line 17: is extra space here intended?

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
+const { CovalentClient } = require('@covalenthq/client-sdk');
+
+// Same addresses as the ones in constants.ts, they can't be imported here because this script is run by node.js
+const ETH_PERSONAL_KEY_ADDRESS = '0xf64750f13f75fb9e2f4d9fd98ab72d742d1e33eb';
+const BNB_DOMAIN_KEY_ADDRESS =  '0x81d205120a9f04d3f1ce733c5ed0a0bc66714c71';
```
hcho112, line 5: wonder if we should move this to secret for github action and for local dev, we should define them inside `.env`?

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
...
+
+function fromWei(wei) {
+  return Number(BigInt(wei)) / 1e18;
+}
```
hcho112, line 9: there is exact same function exist on `ethers` on `near-fast-auth-signer/package.json` where:
```
const { ethers } = require('ethers');

function fromWei(wei) {
  return ethers.utils.formatEther(wei);
}
```
wonder if we should just import existing one?

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
...
+    address
+  );
+
+  if (response.error) {
+    throw new Error(`Failed to fetch balance: ${response.error_message}`);
+  }
+
+  const { balance } = response.data.items[0];
+  if (balance === null) return 0;
+  // type: Response<BalancesResponse>
```
hcho112, line 26: clean up if not needed

packages/near-fast-auth-signer-e2e-tests/testAccounts.json
```diff
@@ -1,4 +1,154 @@
...
+    {
+      "publicKey": "ed25519:3t2g6amQf2WW3XaKdHzYoQgzfUq3LD3VymFeHk32mNso",
+      "accountId": "testuser-playwright-fastauth-13-1718836671240.testnet"
+    },
+    {
+      "publicKey": "ed25519:FSnpBJTv9JRavuqwCkgecZsEtj7Z55U3i8ojrX3jtnrH",
+      "accountId": "testuser-playwright-fastauth-13-1718836671240.testnet"
+    }
+  ]
+}
```
hcho112, line 154: these shouldn't be committed, it should be empty. 
I

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  test.beforeEach(() => {
+    test.slow();
+  });
+
+  const ensureWalletSelectorIsLoaded = async () => {
+    const walletSelector = page.locator('#ws-loaded');
+    await expect(walletSelector).toBeVisible();
+  };
```
hcho112, line 58: there is a util function called `isWalletSelectorLoaded` inside `walletSelector.ts` that can replace this

packages/near-fast-auth-signer-e2e-tests/utils/constants.ts
```diff
@@ -3,3 +3,17 @@ import { Page } from 'playwright/test';
...
 export const TIMEOUT = 30000;
+
+// Below are the static derived addresses
+export const ETH_PERSONAL_KEY_ADDRESS = '0xf64750f13f75fb9e2f4d9fd98ab72d742d1e33eb';
+export const BNB_DOMAIN_KEY_ADDRESS =  '0x81d205120a9f04d3f1ce733c5ed0a0bc66714c71';
+
+// BNB_PERSONAL_KEY_ADDRESS => '0xf64750f13f75fb9e2f4d9fd98ab72d742d1e33eb';
+// ETH_UNKNOWN_KEY_ADDRESS => '0xf64750f13f75fb9e2f4d9fd98ab72d742d1e33eb';
+// BNB_UNKNOWN_KEY_ADDRESS => '0xf64750f13f75fb9e2f4d9fd98ab72d742d1e33eb';
+// ETH_DOMAIN_KEY_ADDRESS =>  '0x81d205120a9f04d3f1ce733c5ed0a0bc66714c71';
```
hcho112, line 14: clean up if not needed or leave a comment

packages/near-fast-auth-signer-e2e-tests/utils/multiChain.ts
```diff
@@ -0,0 +1,109 @@
+import * as bitcoinLib from 'bitcoinjs-lib';
+import canonicalize from 'canonicalize';
+import { fetchDerivedEVMAddress, fetchDerivedBTCAddressAndPublicKey } from 'multichain-tools';
+
+export function toWei(eth: number): string {
+  const wei = eth * 1e18;
+  return wei.toString();
+}
```
hcho112, line 8: there is existing function on ethers with

```
export function toWei(eth: number): string {
  const wei = eth * 1e18;
  return wei.toString();
}
```

packages/near-fast-auth-signer-e2e-tests/utils/multiChain.ts
```diff
@@ -0,0 +1,109 @@
...
+      const { address: btcAddress } = await fetchDerivedBTCAddressAndPublicKey(
+        accountId,
+        derivationPath,
+        bitcoinLib.networks.testnet,
+        'testnet',
+        'v2.multichain-mpc.testnet',
+      );
+      address = btcAddress;
+    } else if (chainValue === 60) {
+      // address = await fetchDerivedEVMAddress(accountId, derivationPath, 'testnet', 'multichain-testnet-2.testnet');
```
hcho112, line 63: delete?

packages/near-fast-auth-signer-e2e-tests/firebase.ts
benmalcom: @hcho112 you're right, I see it's useless now, i've removed it.

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
benmalcom: I'm indifferent about this.
But I think we should leave it here,  we can quickly see the use of the addresses from the comment on top of it instead of figuring out what it does an an `.env ` variable.

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
...
+  if (!response.ok) {
+    return response.text().then((text) => { throw new Error(text); });
+  }
+
+  console.log(`Testnet token request for ${chain} address ${address} successful`);
+
+  return response.json();
+}
+
+async function requestTokensFromFaucet() {
```
Pessina, line 68: Shouldn't we make the `address/chain` arguments of the function instead of hard code taht we will fetch only for ETH/BNB? 

So, in the future, will be easier to add test for more chains

packages/near-fast-auth-signer-e2e-tests/test-app/components/SignMultiChain.tsx
```diff
@@ -0,0 +1,166 @@
...
+  onSubmitForm: (values: {
+    keyType: string,
+    chainValue: number,
+    amount: number,
+    chainId: string,
+    address: string}) => void
+}
+export default function SignMultiChain(props: SignMultiChainProps) {
+  const { onSubmitForm } = props;
+  const onSubmit = (event: FormEvent<HTMLFormElement>) => {
```
Pessina, line 14: I think we are using `react-hook-form` and `yup` for form management on `fast-auth-signer`, no? 

If so, can you please re-use it here to make it consistent with the codebase?

packages/near-fast-auth-signer-e2e-tests/tests/sign/sign-transaction.spec.ts
```diff
@@ -116,6 +116,7 @@ describe('Sign transaction', () => {
       networkId: 'testnet',
       provider:  { type: 'JsonRpcProvider', args: { url: 'https://rpc.testnet.near.org' } },
       signer:    { type: 'InMemorySigner', keyStore: new InMemoryKeyStore() },
+      // @ts-ignore
```
Pessina, line 119: Plase explain the @ts-ignore

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+test.describe('Sign MultiChain', () => {
+  test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext();
+    page = await context.newPage();
+    signMultiChain = new SignMultiChain(page);
+    await page.goto('/');
+    await page.waitForLoadState('domcontentloaded');
+    // Disable pointer events that will interrupt click event on "Approve button"
```
Pessina, line 41: Explain why this is necessary. This doesn't seems to be something that the user can do on his side while using the app

Pessina (changes requested):

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
...
+  if (response.error) {
+    throw new Error(`Failed to fetch balance: ${response.error_message}`);
+  }
+
+  const { balance } = response.data.items[0];
+  if (balance === null) return 0;
+  return fromWei(BigInt(balance).toString());
+}
+
+function getWalletChain(chain) {
```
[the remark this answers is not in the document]
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
```diff
@@ -0,0 +1,87 @@
...
+    case 'eth':
+      return 'sepolia';
+    case 'bnb':
+      return 'bnb-testnet';
+    default:
+      return null;
+  }
+}
+
+async function fillWallet(address, chain) {
```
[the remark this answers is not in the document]
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/test-app/components/SignMultiChain.tsx
```diff
@@ -0,0 +1,166 @@
+import React, { FormEvent } from 'react';
+
+type SignMultiChainProps = {
+  // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/test-app/components/SignMultiChain.tsx
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/tests/sign/sign-transaction.spec.ts
benmalcom: Removed.
It stops my tests from running as there's unimplemented with error `Property useLocalViewExecution is missing in type`, had to disable the warning before running mine, guess it's not so for every other person.

packages/near-fast-auth-signer-e2e-tests/scripts/faucet.js
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
benmalcom: `webpack-dev-server` adds a hidden iframe to projects (for hot reloading), that iframe stays on top of the page components and causes event propagation, which means a click event by playwright will have multiple effects, so it needs to be disbaled.

This can't happen in `prod` for ideal user because `webpack-dev-server` is used for development.

Pessina: Got it, please add this explanation on the code so it's easier for us to know why we are disabling it.

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
Pessina: Please add details to the ticket: https://github.com/near/fast-auth-signer/issues/259

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
benmalcom: Done

packages/near-fast-auth-signer-e2e-tests/tests/signMultiChain/sign-multi-chain.spec.ts
benmalcom: Done

Pessina (approved):

Pessina:
@benmalcom thanks for fixing all the comments! LGTM!

benmalcom:
@hcho112 can you approve this as the last change from [this PR](https://github.com/near/fast-auth-signer/pull/260) dismissed @Pessina review.

.github/workflows/playwright.yml
```diff
@@ -59,18 +59,29 @@ jobs:
...
         cd ${{ github.workspace }}/packages/near-fast-auth-signer-e2e-tests
         npm install -g yarn && yarn
     - name: Install Playwright Browsers
       run: yarn playwright install --with-deps
+    # This .env variables will be available to be used in the signer app
+    - name: Send secrets to process.env
+      run: |
+        echo "ETH_PROVIDER_URL_TESTNET=${{ secrets.ETH_PROVIDER_URL_TESTNET }}" >> $GITHUB_ENV
+        echo "BNB_PROVIDER_URL_TESTNET=${{ secrets.BNB_PROVIDER_URL_TESTNET }}" >> $GITHUB_ENV
+        echo "BTC_PROVIDER_URL_TESTNET=${{ secrets.BTC_PROVIDER_URL_TESTNET }}" >> $GITHUB_ENV
```
hcho112, line 74: why does it have to be passed down to .env instead of directly feed into process.env like 
`MULTICHAIN_TEST_ACCOUNT_ID: ${{ secrets.MULTICHAIN_TEST_ACCOUNT_ID }}`

hcho112 (approved):
Left one comment for clarification but rest looks good

.github/workflows/playwright.yml
benmalcom: @hcho112 that's right. 
Fixed.

hcho112 (approved):
LGTM

PR #248, opened 2024-06-10
Replacing hardcode RPC URLs with environment variables

hcho112:
This PR contains implementation to replace hardcoded URL to environment variables. 
#243


hcho112:
I wasn't sure if we need to also support receiving provider URL via query param. @Pessina let me know what do you think?

Pessina (approved):

Pessina (approved):

PR #252, opened 2024-06-11
Add transactionHash to multiChain response post message

benmalcom:
Include `transactionHash` as part of `postMessage` to parent to signal a successful transaction.


esaminu (approved):
LGTM

PR #260, opened 2024-06-21
Add env variables to webpack plugin

benmalcom:
`.env` variables were not added to webpack environment plugin which means they show as `undefined`.
This PR adds it.

### DIFF
--- a/packages/near-fast-auth-signer/webpack.common.config.js
+++ b/packages/near-fast-auth-signer/webpack.common.config.js
@@ -22,8 +22,6 @@ const envVars = Object.keys(process.env).reduce((prev, next) => {
   return prev;
 }, {});
 
-console.log('envVars ', envVars);
-
 module.exports =      {
   entry:  './src/index.tsx',
   output: {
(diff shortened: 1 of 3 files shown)


packages/near-fast-auth-signer/webpack.common.config.js
```diff
@@ -1,12 +1,23 @@
...
 const { sentryWebpackPlugin } = require('@sentry/webpack-plugin');
+const dotenv = require('dotenv');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const webpack = require('webpack');
 
 // Execute Git command to get the current commit ID
 const commitHash = childProcess.execSync('git rev-parse --short HEAD').toString().trim();
+// Load environment variables from .env file or process.env
+const envFilePath = path.resolve(__dirname, '.env');
+const envConfig = fs.existsSync(envFilePath) ? dotenv.parse(fs.readFileSync(envFilePath)) : process.env;
```
hcho112, line 14: With this approach, if there is .env and also env variables being pass through commend line, would env from commend line gets ignored?

benmalcom: Hey @hcho112 nope it will not get ignored, it will be included.
And it will also take precedence.

hcho112 (approved):

PR #263, opened 2024-06-27
Fix/multichain test browser rendering

benmalcom:
The submit button on [this line](https://github.com/near/fast-auth-signer/blob/6eea2b57b3c3087a8ddb7dd965a6198adf066dc7/packages/near-fast-auth-signer-e2e-tests/tests/models/SignMultiChain.ts#L50) does not work on first click in webkit browsers, so the transaction does not even submit for the test to run. Until it double clicks.
The quick fix now is to double click the submit button if the browser is webkit, else traditional click.

### DIFF
--- a/packages/near-fast-auth-signer-e2e-tests/tests/models/SignMultiChain.ts
+++ b/packages/near-fast-auth-signer-e2e-tests/tests/models/SignMultiChain.ts
@@ -22,8 +22,11 @@ type TransactionDetail = {
 class SignMultiChain {
   private readonly page: Page;
 
+  private isWebkit: boolean;
+
   constructor(page: Page) {
     this.page = page;
+    this.isWebkit = false;
   }
 
   // Wait for about 4 minutes to get multichain response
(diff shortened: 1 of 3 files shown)


packages/near-fast-auth-signer-e2e-tests/tests/models/SignMultiChain.ts
```diff
@@ -22,8 +22,11 @@ type TransactionDetail = {
 class SignMultiChain {
```
Pessina, line 22: I think this is not the right solution, we can't expect the user to double click it. 

I was investigating this and looks like the error comes from the forms validation, as you can see on the video attached. I'm checking how to fix it. 

Idk why it only happens in Safari... very weird. 

https://github.com/user-attachments/assets/1501f7fd-ede8-481a-8654-317dfd080870

Pessina: I did some changes on the forms and now it's passing locally with a single click (have to test on the CI yet)

The failing tests are the new ones I introduced. I'm working on it. 

I will push the fix on my PR for the EVM function call. 

Screenshot 2024-07-19 at 15 28 48

Pessina (changes requested):

packages/near-fast-auth-signer-e2e-tests/tests/models/SignMultiChain.ts
benmalcom: Hey @Pessina thanks for spotting that, looking forward to your solution. Although the problem here is just on the local form we setup for testing e2e, that is, triggering the local form input click with playwright, we don't have such problem on real implementations like the demo app.

Pessina: Yeah, agree, it's a problem only on the e2e app.

But it's better to fix the root problem instead of a quick fix. 

I will open the PR soon, I'm struggling a bit with my tests now haha

benmalcom: Haha okay then.
Do you want me to close this PR then?

Pessina: Up to you. 

Do you wanna check my solution first, to see if you are happy with it?

benmalcom: All right then. 
Let me wait.

PR #264, opened 2024-07-05
Change auth flow if account does not exist during sign in

benmalcom:
This PR changes the authentication flow when a user without an account tends to sign in, once we can't find the account of this user in the `AuthCallbackPage`, we show the `CreateAccountForm` and populate it with their intended details.

Follow this I extracted the create account process in to a form to be reused and performed some other related refactors.


hcho112:
@benmalcom I just tried on this PR locally and it is working as expected. Only thing that I would like to see as improvement is that is it possible to directly create an account from below page to next instead of double verification on email?
image

benmalcom:
@hcho112 if I get what you're saying, you think it will be okay to directly create an account since they're already coming to this page from their inbox which proves they own the email right?

hcho112:
@benmalcom yes! Initially I was little concern about this might expose email being used. However, I realise that in order to get to next step, they need to verify via given email first so we don't have to worry about security. If so, next thing came to my mind is UX there from user perspective, they already proven that they are the owner of the email, therefore we should allow them to create account without double verifying it

benmalcom:
@esaminu what do you think?

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -1,290 +1,164 @@
...
 
-  if (callbackError) return <AuthCallbackError error={callbackError} redirectUrl={decodeIfTruthy(searchParams.get('failure_url'))} />;
-
-  return <StyledStatusMessage data-test-id="callback-status-message">{statusMessage}</StyledStatusMessage>;
+  if (!isAccountExisting) {
+    return (
+      <PageWrap>
+        <AccountInfo>Oops! This account doesn&apos;t seem to exist. Please create one below.</AccountInfo>
+        <CreateAccountForm
+          onSubmit={createAccount}
```
esaminu, line 142: Yes @benmalcom we already have the firebase token in this scope so no need to resend the email again we just need to prompt the user to choose an accountId and talk to mpc to create it

benmalcom:
Fixed

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -1,290 +1,193 @@
...
+  border-radius: 4px;
+  border: 1px solid #FBD38D;
+  background: #FEEBC8;
+  box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.06);
+  @media (min-width: 768px) {
+    max-width: 380px;
+  }
+  margin-bottom: 12px;
+  color: #2D3748;
+  p{
```
hcho112, line 44: extra spacing needed

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -1,290 +1,193 @@
...
-              accountId:    accountIds[0]
-            }
-          }, '*');
-        } else {
-          window.location.replace(parsedUrl.href);
-        }
-      }
-    });
-};
+const StatusMessage = styled.p`
```
hcho112, line 49: if we do not need custom css, should we just use default `p` element ?

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
```diff
@@ -1,290 +1,193 @@
...
     const signInProcess = async () => {
-      if (isSignInWithEmailLink(firebaseAuth, window.location.href)) {
-        const accountId = decodeIfTruthy(searchParams.get('accountId'));
-        const isRecovery = decodeIfTruthy(searchParams.get('isRecovery'));
-        const success_url = decodeIfTruthy(searchParams.get('success_url'));
-        const public_key_lak = decodeIfTruthy(searchParams.get('public_key_lak'));
-        const contract_id = decodeIfTruthy(searchParams.get('contract_id'));
-        const methodNames = decodeIfTruthy(searchParams.get('methodNames'));
+      if (!isSignInWithEmailLink(firebaseAuth, window.location.href)) {
+        navigate('/signup');
```
hcho112, line 65: would this navigate inherit all query params?

packages/near-fast-auth-signer/src/components/Layout/FormContainer.tsx
```diff
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 
 const FormContainer = styled.form<{ inIframe?: boolean }>`
   width: 375px;
-  margin: 16px auto;
+  //margin: 16px auto;
```
hcho112, line 5: remove

hcho112 (changes requested):
Left few comments. 

Since this actually changes create account flow, would be nicer if we can have test to cover this?

hcho112:
Also tests are failing, would it be to do with flaky? or actual failure?

packages/near-fast-auth-signer/src/components/AuthCallback/AuthCallback.tsx
benmalcom: We don't need it because it is going to the signup page of the gateway directly.

hcho112:
I actually manually tested this PR and it seems to work just fine. Should we merge it first or wait for test to be recovered?

hcho112 (dismissed):
Again, I have re-tested both create and sign in flow and they all work as expected, since test outage is not to do with this PR, I will approve this

Pessina (approved):

PR #265, opened 2024-07-17
feat: add cloud functions

esaminu:


packages/functions/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+});
+
+export const sendOTP = functions.https.onCall(
+  async (data: { email: string }, _context) => {
+    const { email } = data;
+
+    const otp = crypto.randomInt(100000, 999999).toString();
+
+    const otpDoc = admin.firestore().collection('otps').doc(email);
+    await otpDoc.set({
```
benmalcom, line 33: As an enhancement, could we add a check to see if there is an existing non-expired OTP for the same email before creating a new one? This would help prevent flooding the otps collection with multiple OTP requests and ensure that old OTPs are not left without cleanup.

packages/functions/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        new Date(Date.now() + 10 * 60 * 1000)
+      ), // 10 minutes expiration
+    });
+
+    const now = new Date();
+    const formattedDate = now.toISOString().replace(/T/, ' ').replace(/\..+/, ' Z');
+
+    const emailContent = `Hello,<br><br>We received a request to sign in to NEAR Onboarding using this email address, at ${formattedDate}. If you want to sign in with your ${email} account, use this code:<br><br>${otp}<br><br>If you did not request this link, you can safely ignore this email.<br><br>Thanks,<br><br>Your NEAR Onboarding team`;
+
+    await transporter.sendMail({
```
benmalcom, line 46: Hey @esaminu, what happens to the  otp `doc` if email sending fails? Or can we store it only when email sending succeeds?

packages/functions/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        'deadline-exceeded',
+        'OTP has expired'
+      );
+    }
+
+    let uid: string;
+    try {
+      const userRecord = await admin.auth().getUserByEmail(email);
+      uid = userRecord.uid;
+    } catch (error) {
```
benmalcom, line 88: Can we make sure it is only a `auth/user-not-found` error that creates a new user?

.firebaserc
```diff
@@ -0,0 +1,5 @@
+{
```
Pessina, line 1: Should we add a cron job to clean up old OTPs?

esaminu: Right now the otp doc that is verified is deleted. I will change this to delete all otps for that email address once one is verified and in addition to that we can add a daily cron job to delete expired otps

packages/functions/src/index.ts
esaminu: There might be some issue with the email receiver that caused it to not receive the previous otp causing the user to click resend. I don't think it's good practice to re use the same code in these cases

.firebaserc
esaminu: Old otps that are left in our db will happen when someone requests codes to multiple emails and does not verify any of them. We create one record per email (we override the record for multiple requests to the same email) and we delete the record once the user is verified and gets a token. We can add a daily cron job to handle expired otps older than a month or so IMO but they serve as a log for people trying to login and unable to get access, we can inspect the db and verify that we sent them a code - we did something similar for wallet

Pessina: Yeah, I was thinking about the expired ones. But it makes sense to leave it there for logging purpose.

Pessina (approved):

Pessina:
Great improvement!!! @esaminu

PR #274, opened 2024-08-06
Social login url fix by including basepath

hcho112:
This PR contains minor fix for social login redirect url. 
because it is missing `basePath` which it doesn't redirect to correct fastauth url


benmalcom (approved):

hcho112:
this test locally passes but fail on CI for some reason. Will investigate later but for the sake of fixing the social login, will merge this first

